PR #178, opened 2026-06-03
feat(app): show ends-at time in check-in toast expanded view

seferturan:
Mirrors the change from trakt-web/pull/2381.

The expiresAt instant is
formatted as a locale-aware short time and displayed below the remaining
duration in the expanded check-in toast.


michaldrabik:
@seferturan Updated notification display and added option to tap "Ends at" to see minutes left optionally.

image

---

image

image

PR #291, opened 2026-07-21
feat: TV syncs and bugfixes

michaldrabik:


michaldrabik (commented):
## Code Review

This pull request delivers TV sync support alongside a batch of bugfixes. The bulk of the change is a structural refactor: shared collaborators (`ProgressItem`, `CollectionStateProvider`, `UserProgressStorage` / `UserProgressLocalDataSource`, and the `recordError` extension) move from `:app` into `:common` so the `:tv` sync flows can reuse them, and the watchlist sync clients swap their loosely-typed `sort: String` parameter for the typed `Sorting` domain type, wiring the API's `sortBy` / `sortHow` fields off `sorting.type.value` / `sorting.order.value`. The remaining ~300 files are one-line import adjustments following those moves, plus a visual refresh of the TV media cards.

The change reads as disciplined. The module moves respect the `app -> common -> resources` dependency direction, the `sort: String` to `Sorting` swap is a genuine improvement in line with the typed-domain principle, the new `:common` model stays `@Immutable`, and the TV media card artwork uses `TraktTheme.colors.*` tokens throughout, with raw `Color(...)` values confined to decorative gradient scrims and preview code, which the theming rules permit. No inline findings were raised against the diff. One item outside the code is worth a quick fix before merge: the PR title `feat: TV syncs and bugfixes` is missing a scope. Per the commit standards, PR titles follow Conventional Commits with the android-scoped enum, so `feat(tv): TV syncs and bugfixes` (or `feat(app):`) would keep the squashed history consistent.

PR #299, opened 2026-07-23
feat: introduce release type for releases/calendar

michaldrabik:
Preview:

[Screen_recording_20260723_102218.webm](https://github.com/user-attachments/assets/56420fc0-14eb-4739-8b5e-58ae86c1b7c4)


app/src/main/java/tv/trakt/trakt/core/home/sections/upcoming/usecases/GetUpcomingUseCase.kt
```diff
@@ -138,14 +144,30 @@ internal class GetUpcomingUseCase(
...
+    // Release type applies to episodes only; Premiere/Finale keep grouped episode
+    // items containing a matching episode and drop movies entirely.
+    private fun List<CalendarItem>.filterByType(type: ReleaseType): List<CalendarItem> {
+        if (type == ReleaseType.All) {
+            return this
+        }
+        return filter { item ->
+            item is CalendarItem.EpisodeItem &&
+                item.episodes.any { episode ->
+                    when (type) {
```
michaldrabik, line 166: high

The `when (type)` here is the return value of the `any { … }` lambda, so it is used as an expression and must be exhaustive over `ReleaseType`. Only `Premiere` and `Finale` are covered, so `ReleaseType.All` is missing and the function will not compile. The early `return this` guard above does not narrow the enum for the compiler. Since `All` is already handled by the guard, we can add an explicit unreachable branch rather than an `else ->` so future variants are still surfaced.

```kotlin
when (type) {
    ReleaseType.All -> false
    ReleaseType.Premiere -> episode.type?.isPremiere == true
    ReleaseType.Finale -> episode.type?.isFinale == true
}
```

app/src/main/java/tv/trakt/trakt/core/discover/sections/releases/usecases/shows/DefaultGetReleasesShowsUseCase.kt
```diff
@@ -34,14 +41,34 @@ internal class DefaultGetReleasesShowsUseCase(
...
-        )
-            .filter { (it.episode?.season ?: 0) > 0 }
-            .map { Show.fromDto(it.show!!) to Episode.fromDto(it.episode!!) }
+        val data = when (type) {
+            All -> remoteSource.getReleases(
+                startDate = startDate,
+                days = days,
+                filters = filters,
+            ).map {
+                Show.fromDto(it.show!!) to Episode.fromDto(it.episode!!)
```
michaldrabik: high

The `All` branch keeps `it.show!!` and `it.episode!!`, and `!!` non-null assertions are not allowed in production code. The `Premiere` and `Finale` branches below already dereference `it.show` / `it.episode` safely because their DTOs are non-null, so the `All` branch is the odd one out and will crash if either field is ever null. We should drop the items with a missing show or episode instead of asserting.

```kotlin
All -> remoteSource.getReleases(
    startDate = startDate,
    days = days,
    filters = filters,
).mapNotNull {
    val show = it.show ?: return@mapNotNull null
    val episode = it.episode ?: return@mapNotNull null
    Show.fromDto(show) to Episode.fromDto(episode)
}
```

app/src/main/java/tv/trakt/trakt/core/home/sections/upcoming/views/EpisodeUpcomingItemView.kt
```diff
@@ -50,16 +49,13 @@ internal fun EpisodeUpcomingItemView(
...
+                    item.episode.isFinale(isLatestAired = midReleases) -> FinaleChip(modifier = shadowModifier)
                 }
 
                 InfoChip(
-                    text = item.releasedAt.toLocal().relativeDateTimeString(),
-                    iconPainter = when {
-                        item.episode.isReleased -> painterResource(R.drawable.ic_calendar_check)
-                        else -> painterResource(R.drawable.ic_calendar_upcoming)
-                    },
+                    text = item.releasedAt?.toLocal()?.relativeDateTimeString() ?: "N/A",
```
michaldrabik, line 57: medium

Now that `releasedAt` is nullable, the fallback text is a hardcoded `"N/A"` literal. User-facing strings in composables should not be inlined - they bypass localization and ship untranslated. We should move this to a namespaced string resource in `values/strings.xml` and read it via `stringResource`.

```kotlin
text = item.releasedAt?.toLocal()?.relativeDateTimeString()
    ?: stringResource(R.string.common_not_available),
```

app/src/main/java/tv/trakt/trakt/core/home/sections/upcoming/views/MovieUpcomingItemView.kt
```diff
@@ -35,11 +33,8 @@ internal fun MovieUpcomingItemView(
         onClick = { onClick(item.movie.ids.trakt) },
         cardContent = {
             InfoChip(
-                text = item.releasedAt.toLocal().relativeDateTimeString(),
-                iconPainter = when {
-                    item.movie.isReleased -> painterResource(R.drawable.ic_calendar_check)
-                    else -> painterResource(R.drawable.ic_calendar_upcoming)
-                },
+                text = item.releasedAt?.toLocal()?.relativeDateTimeString() ?: "N/A",
```
michaldrabik, line 36: medium

Same hardcoded `"N/A"` fallback as in `EpisodeUpcomingItemView`. This user-facing text should come from a localized string resource rather than an inline literal so it translates through Crowdin.

```kotlin
text = item.releasedAt?.toLocal()?.relativeDateTimeString()
    ?: stringResource(R.string.common_not_available),
```

app/src/main/java/tv/trakt/trakt/core/summary/episodes/data/EpisodeDetailsUpdates.kt
```diff
@@ -10,9 +10,9 @@ internal interface EpisodeDetailsUpdates {
 
     enum class Source {
         HOME,
-        PROGRESS,
-        HISTORY,
-        SEASON,
+        Progress,
+        History,
+        Season,
         CALENDAR,
```
michaldrabik: low

This migration renames `PROGRESS` / `HISTORY` / `SEASON` to PascalCase but leaves `HOME` and `CALENDAR` in SCREAMING_SNAKE_CASE, so the enum now mixes both conventions. Enum entries follow PascalCase (SCREAMING_SNAKE is reserved for `const val`). We should rename the remaining two for consistency and update their call sites.

```kotlin
enum class Source {
    Home,
    Progress,
    History,
    Season,
    Calendar,
}
```

michaldrabik (commented):
## Code Review

This pull request introduces a release-type filter (All / Premiere / Finale) across the calendar, discover releases, all-releases, and home upcoming surfaces, backed by new `GetCalendarTypeUseCase` / `GetReleasesTypeUseCase` DataStore stores and dedicated premiere/finale show endpoints. It also unifies the home upcoming section onto the shared `CalendarItem` model, adds `clearLocal()` to the releases caches so a type change does not serve stale items, and migrates several `delay` / `debounce` calls to typed `Duration`.

Feedback on the changes focuses on a non-exhaustive `when` over `ReleaseType` that omits the `All` case, a retained `!!` pair on the `All` releases branch, a couple of hardcoded `"N/A"` fallbacks that bypass localization, and a half-migrated enum whose entry casing is now inconsistent.

PR #306, opened 2026-07-27
feat: comments reports

michaldrabik:
image

image

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/core/comments/features/report/ReportCommentSheet.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/comments/features/report/ReportCommentSheet.kt
@@ -1,0 +1,92 @@
+@file:OptIn(ExperimentalMaterial3Api::class)
+
+package tv.trakt.trakt.core.comments.features.report
+
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.SheetState
+import androidx.compose.material3.rememberModalBottomSheetState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalResources
+import androidx.compose.ui.unit.dp
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import org.koin.androidx.compose.koinViewModel
+import org.koin.core.parameter.parametersOf
+import tv.trakt.trakt.LocalSnackbarState
+import tv.trakt.trakt.common.model.Comment
+import tv.trakt.trakt.resources.R
+import tv.trakt.trakt.ui.components.TraktBottomSheet
+import tv.trakt.trakt.ui.snackbar.SNACK_DURATION_SHORT
+import kotlin.random.Random
+import kotlin.time.Duration.Companion.milliseconds
+
+@Composable
+internal fun ReportCommentSheet(
+    state: SheetState = rememberModalBottomSheetState(
+        skipPartiallyExpanded = true,
+    ),
+    active: Boolean,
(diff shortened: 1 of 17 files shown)


app/src/main/java/tv/trakt/trakt/core/comments/features/report/ReportCommentSheet.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    val sheetScope = rememberCoroutineScope()
+
+    if (active && comment != null) {
+        TraktBottomSheet(
+            sheetState = state,
+            onDismiss = onDismiss,
+        ) {
+            ReportCommentView(
+                viewModel = koinViewModel(
```
michaldrabik, line 48: high

`koinViewModel(key = Random.nextInt().toString())` recomputes the key on every recomposition, because `Random.nextInt()` runs each time this composable executes. Since the key identifies the ViewModel in the store, a fresh key means a brand new `ReportCommentViewModel` is created on every recomposition. When the user submits, `report()` flips the loading state, which triggers a recomposition, which spawns a new ViewModel whose `state` is back at `Idle` - the in-flight coroutine runs on the discarded instance, so `reported` is never observed, `onReported` never fires, and the sheet neither dismisses nor shows the confirmation. This is the same "keys must be stable across recompositions" pitfall the styleguide calls out for image requests. To get a fresh instance per sheet opening while staying stable across recompositions, we should remember the key:

```kotlin
val viewModelKey = remember(comment) { comment.id.toString() }
// ...
viewModel = koinViewModel(
    key = viewModelKey,
    parameters = { parametersOf(comment) },
),
```

app/src/main/java/tv/trakt/trakt/core/comments/features/report/ReportCommentSheet.kt
```diff
@@ -0,0 +1,92 @@
...
+                    parameters = {
+                        parametersOf(comment)
+                    },
+                ),
+                onReported = {
+                    onReported()
+                    sheetScope.dismissWithAction(
+                        sheet = state,
+                        onDismiss = onDismiss,
+                        action = {
```
michaldrabik, line 59: medium

The success snackbar is launched on `sheetScope` (a `rememberCoroutineScope()` bound to this composable), but it runs from inside `dismissWithAction`, which calls `onDismiss()` immediately after - flipping `reportActive` to false and removing the sheet from composition. When the sheet leaves composition its `sheetScope` is cancelled, cancelling the `showSnackbar` coroutine, so the confirmation can be dismissed the moment it appears. The manual `launch` / `delay` / `job.cancel()` dance is fighting this teardown rather than avoiding it. The styleguide asks that one-shot events like snackbars flow through a `SharedFlow(replay = 0)` collected in a `LaunchedEffect`, or at minimum be shown from a scope that outlives the sheet. We can lift the message out of the sheet's scope - for example collect a `reported` event in a `LaunchedEffect` at the call site, or launch the snackbar on an app-level scope - so it survives dismissal:

```kotlin
LaunchedEffect(Unit) {
    viewModel.events.collect { localSnack.showSnackbar(message) }
}
```

michaldrabik (commented):
## Code Review

This pull request adds the ability to report comments and replies. It introduces a `ReportCommentSheet` bottom sheet backed by a `ReportCommentViewModel`, a `ReportReason` enum, a `ReportCommentUseCase`, and a reusable `CommentDropdown` overflow menu wired into both `CommentCard` and `CommentReplyCard`. The data layer gains a `postReport` call on `CommentsRemoteDataSource` / `CommentsApiClient`, plus a `429` rate-limit constant, a flag icon, and the associated strings.

Feedback on the changes focuses on the lifecycle of the report sheet: the `ReportCommentViewModel` is keyed with a value that is recomputed on every recomposition, which recreates the ViewModel and breaks the report flow, and the success snackbar is dispatched from a coroutine scope that is torn down as the sheet dismisses, so the confirmation may not display reliably.

PR #308, opened 2026-07-27
fix: private followed users access

michaldrabik:
- navigating to self should not open "external user" profile.
- display user private profile if you are following them

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/core/main/MainScreen.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/main/MainScreen.kt
@@ -160,7 +160,7 @@ internal fun MainScreen(
 
     LaunchedUpdateEffect(state.user) {
         when {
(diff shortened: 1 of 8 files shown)


michaldrabik (commented):
## Code Review

This pull request fixes access to private profiles for users the viewer already follows, and prevents navigating to one's own profile. It threads the current user's id through the main navigation graph so `navigateToUserProfile` can short-circuit self-navigation, renames `MainState.loadingUser` to `userLoading`, and reworks `UserProfileViewModel` so private-but-followed profiles load full content while a loading indicator is shown until following status resolves.

This review checked the changes against the repository style guide and did not find any violations to raise. The supporting refactors - the shared `CoroutineScope.loadProfileContent` helper, the flattened private-profile branch in `UserProfileContent`, and the method-reference navigation lambdas - are consistent with the existing conventions in the codebase.

PR #311, opened 2026-07-28
feat: a bit more sections syncs for TV

michaldrabik:


tv/src/main/java/tv/trakt/trakt/app/core/profile/sections/favorites/ProfileFavoritesViewModel.kt
```diff
@@ -74,15 +99,17 @@ internal class ProfileFavoriteMoviesViewModel(
...
-    ) { s1, s2, s3 ->
-        ProfileFavoriteMoviesState(
-            isLoading = s1,
-            items = s2,
-            error = s3,
+    ) { loading, items, _, error ->
+        ProfileFavoritesState(
+            isLoading = loading,
+            items = items,
+            collection = UserCollectionState.Default,
```
michaldrabik: high

The `combine` here collects `collectionStateProvider.stateFlow` as its third source but discards it (`_`) and hardcodes `collection = UserCollectionState.Default` in the emitted state. Because `ProfileFavoritesView` reads `state.collection.isWatched(...)`, `isWatching(...)` and `isWatchlist(...)` to drive the card badges, the watched/watching/watchlist indicators on the Favorites section will always reflect an empty collection and never update as the user's collection changes. To fix this, we should bind the collected value into the state rather than substituting the default.

```kotlin
    ) { loading, items, collection, error ->
        ProfileFavoritesState(
            isLoading = loading,
            items = items,
            collection = collection,
            error = error,
        )
    }.stateIn(
```

tv/src/main/java/tv/trakt/trakt/app/core/profile/sections/favorites/viewall/ProfileFavoritesViewAllViewModel.kt
```diff
@@ -0,0 +1,90 @@
...
+    val state = combine(
+        loadingState,
+        itemsState,
+        collectionStateProvider.stateFlow,
+        errorState,
+    ) { loading, items, _, error ->
+        ProfileFavoritesViewAllState(
+            isLoading = loading,
+            items = items,
+            collection = UserCollectionState.Default,
```
michaldrabik: high

Same issue as the Favorites section view-model: the `combine` collects `collectionStateProvider.stateFlow` but ignores it (`_`) and emits `collection = UserCollectionState.Default`. The view-all grid passes `state.collection` into `isWatched`/`isWatching`/`isWatchlist` for every card, so the collection badges here will always render as if nothing is collected. We should forward the collected collection into the emitted state.

```kotlin
    ) { loading, items, collection, error ->
        ProfileFavoritesViewAllState(
            isLoading = loading,
            items = items,
            collection = collection,
            error = error,
        )
    }.stateIn(
```

tv/src/main/java/tv/trakt/trakt/app/core/movies/MoviesScreen.kt
```diff
@@ -338,6 +341,112 @@ private fun TrendingMoviesList(
...
+                    HorizontalMediaCard(
+                        title = "",
+                        containerImageUrl = item.images?.getFanartUrl(),
+                        onClick = { onClick(item) },
+                        cardContent = {
+                            Column(
+                                verticalArrangement = spacedBy(2.dp),
+                            ) {
+                                InfoChip(
+                                    text = item.releaseAt?.toLocal()?.relativeDateTimeString() ?: "TBA",
```
michaldrabik, line 396: medium

The `"TBA"` fallback for a missing release date is a raw string literal rendered directly to the user, which will not be translated for any non-English locale. We should move it to a localized string resource in `values/strings.xml` and read it via `stringResource`. The same literal appears in `ShowsScreen.kt:397`, `MoviesReleasesViewAllScreen.kt:138`, and `ShowsReleasesViewAllScreen.kt:145`, so all four sites should point at the shared key.

```kotlin
                                    text = item.releaseAt?.toLocal()?.relativeDateTimeString()
                                        ?: stringResource(R.string.tag_text_release_date_tba),
```

michaldrabik (commented):
## Code Review

This pull request expands the Android TV sync surface: it adds a Recommended row to the Home screen (with a dedicated view-all screen and shared `RecommendedItem` model), replaces the per-media Recommended screens on Shows and Movies with new Releases sections backed by the calendar endpoint, merges the separate favorite-shows and favorite-movies flows into a single interleaved Favorites section and view-all, and reworks the Home upcoming use-case to group same-day episodes into a single card. Several focus-handling refinements (refocus-on-load, exit-direction guards) accompany the new sections.

Feedback on the changes focuses on a state-management defect in the merged Favorites flow, where the collaborator collection state is collected but dropped from the emitted UI state, and on a localization gap where a fallback date label is inlined as a raw string rather than a localized resource.

PR #320, opened 2026-08-03
feat: smart lists (read)

michaldrabik:
- introduces smart lists into the app
- this PR is read only part, creation will be handled in another one


app/src/main/java/tv/trakt/trakt/core/lists/sections/smart/data/local/ListsSmartStorage.kt
```diff
@@ -0,0 +1,58 @@
...
+        mutex.withLock {
+            updatedAt.tryEmit(nowUtcInstant())
+        }
+    }
+
+    override fun observeUpdates(): Flow<Instant?> {
+        return updatedAt.asSharedFlow()
+    }
+
+    override fun clear() {
```
michaldrabik: medium

Every other operation on `storage` acquires `mutex`, but `clear()` mutates the map and emits without taking the lock. If the cache is cleared (e.g. on logout) while a refresh is suspended inside `setLists` on another dispatcher, the unsynchronized mutation can interleave with the locked writers and corrupt the plain `mutableMapOf`. To keep the locking invariant consistent, we should route `clear()` through the same mutex — the sibling liked/collaborations storages share this shape, so they could be aligned in a follow-up.

```kotlin
override suspend fun clear() {
    mutex.withLock {
        storage.clear()
        updatedAt.tryEmit(null)
    }
}
```

app/src/main/java/tv/trakt/trakt/core/lists/sections/smart/details/navigation/SmartListDetailsNavigation.kt
```diff
@@ -0,0 +1,39 @@
...
+import androidx.navigation.compose.composable
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import org.koin.androidx.compose.koinViewModel
+import tv.trakt.trakt.common.model.TraktId
+import tv.trakt.trakt.common.model.lists.SmartList
+import tv.trakt.trakt.core.lists.sections.smart.details.SmartListDetailsScreen
+
+@Serializable
+internal data class SmartListDetailsDestination(
```
michaldrabik, line 14: medium

The route serializes the entire `SmartList` — including all fifteen filter collections and the poster URL list — into a JSON string argument. Navigation arguments end up in the saved instance state `Bundle`, so large payloads count against the binder transaction limit, and a stringly-typed blob loses the benefit of the typed-route convention. Since `GetSmartListsUseCase` already persists lists to `ListsSmartLocalDataSource` before any card is shown, we should pass just the id and let the ViewModel read the list from the local source.

```kotlin
@Serializable
internal data class SmartListDetailsDestination(
    val listId: Long,
)

internal fun NavController.navigateToSmartListDetails(list: SmartList) {
    navigate(route = SmartListDetailsDestination(listId = list.ids.trakt.value))
}
```

app/src/main/java/tv/trakt/trakt/core/lists/sections/smart/usecases/GetSmartListItemsUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+            when {
+                movie != null -> SmartListItem.MovieItem(
+                    rank = dto.rank,
+                    movie = Movie.fromDto(movie),
+                )
+                show != null -> SmartListItem.ShowItem(
+                    rank = dto.rank,
+                    show = Show.fromDto(show),
+                )
+                else -> throw IllegalArgumentException("Smart list item missing media: type=${dto.type}")
```
michaldrabik: medium

When a single item arrives with neither `movie` nor `show` — for example if the API starts returning another media type for `type=all` — the mapping throws and the whole page load fails, surfacing a raw exception message in the snackbar. Following the mapper convention of handling nullability with sensible defaults, we should skip unmappable items instead of failing the entire response.

```kotlin
}.asyncMap { dto ->
    val movie = dto.movie
    val show = dto.show
    when {
        movie != null -> SmartListItem.MovieItem(rank = dto.rank, movie = Movie.fromDto(movie))
        show != null -> SmartListItem.ShowItem(rank = dto.rank, show = Show.fromDto(show))
        else -> {
            Timber.w("Smart list item missing media: type=%s", dto.type)
            null
        }
    }
}.filterNotNull().toImmutableList()
```

app/src/main/java/tv/trakt/trakt/ui/components/mediacards/list/SmartListCard.kt
```diff
@@ -0,0 +1,284 @@
...
+            modifier = Modifier
+                .size(36.dp)
+                .background(
+                    color = TraktTheme.colors.chipContainer,
+                    shape = CircleShape,
+                ),
+        ) {
+            Icon(
+                painter = painterResource(
+                    when (list.mediaType) {
```
michaldrabik: medium

The `when` over `MediaMode` uses an `else ->` branch, so a new enum entry would silently fall into the shows+movies icon instead of failing compilation. `MediaMode` already carries exactly this mapping in its `offIcon` property (`Media -> ic_shows_movies`, `Shows -> ic_shows_off`, `Movies -> ic_movies_off`), so we can drop the duplicated mapping entirely.

```kotlin
Icon(
    painter = painterResource(list.mediaType.offIcon),
    contentDescription = null,
    tint = TraktTheme.colors.textSecondary,
    modifier = Modifier.size(18.dp),
)
```

app/src/main/java/tv/trakt/trakt/ui/components/mediacards/list/SmartListDropdown.kt
```diff
@@ -0,0 +1,83 @@
...
+@Composable
+internal fun SmartListDropdown(
+    onDeleteClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    containerColor: Color = TraktTheme.colors.dialogContainer,
+) {
+    var showMenu by remember { mutableStateOf(false) }
+
+    Box(modifier = modifier) {
+        Icon(
```
michaldrabik, line 38: medium

The dropdown trigger is an icon-only actionable element with `contentDescription = null` and a 16dp clickable area. TalkBack users cannot discover or announce the menu, and the touch target is well below the 48dp minimum. We should give the trigger a content description and let it expand to the minimum interactive size.

```kotlin
Icon(
    painter = painterResource(R.drawable.ic_more_vertical),
    contentDescription = stringResource(R.string.button_text_more),
    tint = TraktTheme.colors.textPrimary,
    modifier = Modifier
        .minimumInteractiveComponentSize()
        .size(16.dp)
        .onClick { showMenu = true },
)
```

common/src/main/java/tv/trakt/trakt/common/model/lists/SmartList.kt
```diff
@@ -0,0 +1,254 @@
...
+    @Serializable(ZonedDateTimeSerializer::class)
+    val createdAt: ZonedDateTime,
+    @Serializable(ZonedDateTimeSerializer::class)
+    val updatedAt: ZonedDateTime,
+    val images: Images?,
+    val source: String,
+    val mediaType: MediaMode,
+    val filters: SmartListFilters,
+) {
+    @Composable
```
michaldrabik, line 46: medium

The domain model in `:common` now owns a `@Composable` function plus `Resources`-based string resolution, `AppCompatDelegate` locale lookup, and `NumberFormat` formatting. Formatting belongs at the edges, and helpers that take `Resources`/`Context` purely to resolve strings are called out in the style guide — models should stay pure data. We should extract `rememberDescription`, `sourceLabel`, and the range formatters into a UI-layer helper (e.g. a `rememberSmartListDescription(list)` composable next to the cards), keeping `SmartList` free of Android presentation dependencies.

```kotlin
// e.g. app/.../ui/components/mediacards/list/SmartListDescription.kt
@Composable
internal fun rememberSmartListDescription(list: SmartList): String { … }
```

app/src/main/java/tv/trakt/trakt/core/lists/sections/smart/details/SmartListDetailsViewModel.kt
```diff
@@ -0,0 +1,246 @@
...
+                        MediaMode.Media -> "all"
+                        MediaMode.Shows -> "shows"
+                        MediaMode.Movies -> "movies"
+                    },
+                    pagination = Pagination(1, PAGE_LIMIT),
+                )
+                    .distinctBy { it.key }
+                    .toImmutableList()
+
+                itemsState
```
michaldrabik: low

`MutableStateFlow.update` returns `Unit`, so the chained `.also { … }` operates on `Unit` and reads as if it were reacting to the updated list. Plain sequential statements say the same thing more honestly.

```kotlin
itemsState.update { items }
hasMoreData = items.size >= PAGE_LIMIT
```

michaldrabik (commented):
## Code Review

This pull request introduces read support for smart lists across the app: a new `SmartList` domain model with filter summaries, a `SmartListsApi`-backed remote data source, an in-memory smart lists cache, a new smart list details screen with pagination and delete support, smart list cards and dropdowns in the Lists and All Lists screens, and a new `ListsItem` sealed type unifying custom and smart lists. It also relocates `CustomList` into `model/lists` and regenerates the OpenAPI client.

Feedback on the changes focuses on thread-safety of the new in-memory storage, resilience of the smart list item mapping, keeping the navigation route payload small and typed, moving presentation concerns out of the `:common` domain model, and a few Compose-side issues around enum exhaustiveness and accessibility of the new dropdown trigger.

PR #322, opened 2026-08-04
fix: fall back to a populated department when known-for has no credits

tysonkerridge:
⚠️ I haven't tried running this code. Example person with this issue: https://app.trakt.tv/people/rob-font

---

Person summary credits (Movies/Shows tabs) could default the department filter to one the person has zero credits in - e.g. `knownForDepartment` says "Acting" but this credits list only has "self" entries - silently rendering an empty row.

- Also fixes the section subtitle hardcoding "Acting" whenever there's exactly one department, even when that department is actually self or narrator.

**Verification:**
- [ ] `:app:compileInternalDebugKotlin` and `:common:compileInternalDebugKotlin` succeed
- [ ] Manually check the Movies and Shows credits rows on a person summary screen, particularly for a person whose only credits are self/narrator appearances, and the subtitle label under "Movies"/"Shows" when there's exactly one department

Refs: [iOS](https://github.com/trakt/trakt-apple-internal/pull/305)/[Web](https://github.com/trakt/trakt-web/pull/3077)

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/core/summary/people/helpers/ResolveSelectedCreditsFilter.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/summary/people/helpers/ResolveSelectedCreditsFilter.kt
@@ -1,11 +1,11 @@
-package tv.trakt.trakt.common.helpers
+package tv.trakt.trakt.core.summary.people.helpers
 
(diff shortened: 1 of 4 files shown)


michaldrabik:
Pending for: trakt/trakt-web#3077

PR #323, opened 2026-08-04
fix: remove 1-year cap on previous-streak lookback

tysonkerridge:
⚠️ I haven't tried running this code

---

`computePreviousStreak` gave up searching for the previous streak's end date once it passed one year before today, so any previous streak that ended further back than that was reported as 0 instead of its real length. This diverged from the iOS and web implementations, which scan the full watch history with no time bound.

The fix bounds the backward search by the earliest day actually present in the activity set instead of a fixed calendar window — removes the artificial cap while still guaranteeing the loop terminates.

**Test plan (needs a real run, not just a read-through)**
- [ ] Build and install a debug build, confirm it compiles clean
- [ ] On an account with a previous streak that ended over a year ago, confirm it now shows the real length instead of 0
- [ ] On an account with no watch history, confirm `previousStreak` still shows 0 (empty-set guard didn't break the normal case)
- [ ] Spot-check a normal account (previous streak within the last year) still matches pre-fix behavior, to confirm nothing regressed for the common case

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/core/home/sections/streaks/data/DefaultStreaksManager.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/home/sections/streaks/data/DefaultStreaksManager.kt
@@ -129,10 +129,14 @@ internal class DefaultStreaksManager(
         today: LocalDate,
         currentStreak: Int,
     ): Int {
+        // Bound the backward search by the earliest real activity instead of a fixed
+        // lookback window, so a previous streak from over a year ago still counts.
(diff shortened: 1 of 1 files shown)


michaldrabik:
Look's fine although I wonder if there is even anyone able to achieve such a streak :P

tysonkerridge:
> Look's fine although I wonder if there is even anyone able to achieve such a streak :P

Sure can 😅 https://forums.trakt.tv/t/daily-streak-activity-differences-between-apps/116311

PR #326, opened 2026-08-05
feat(smart): implement smart list creation

michaldrabik:


common/src/main/java/tv/trakt/trakt/common/model/lists/SmartList.kt
```diff
@@ -161,25 +171,50 @@ data class SmartListFilters(
...
-                imdbRatings = dto.imdbRatings.orEmpty().toImmutableList(),
-                rtMeters = dto.rtMeters.orEmpty().toImmutableList(),
-                rtUserMeters = dto.rtUserMeters.orEmpty().toImmutableList(),
-                ignoreWatched = dto.ignoreWatched ?: false,
-                ignoreWatchlisted = dto.ignoreWatchlisted ?: false,
+                source = SmartListSource.fromApiValue(dto.source),
+                media = when (dto.mediaType) {
+                    "shows" -> MediaMode.Shows
+                    "movies" -> MediaMode.Movies
+                    else -> MediaMode.Media
```
michaldrabik, line 196: high

The `SmartListFilters` constructor requires `media` to be either `Shows` or `Movies` (the `require` in `init`), but this mapping falls back to `MediaMode.Media` for any unrecognised `mediaType`. When the API returns a smart list whose `mediaType` is neither `shows` nor `movies`, `fromDto` constructs `SmartListFilters` with `MediaMode.Media`, which trips that `require` and throws `IllegalArgumentException`. Since `SmartList.fromDto` runs while parsing the user's smart lists, a single unexpected value can crash the whole lists screen. A mapper should coerce unexpected input to a supported default rather than build a value the constructor rejects.

```kotlin
media = when (dto.mediaType) {
    "movies" -> MediaMode.Movies
    else -> MediaMode.Shows
},
```

app/src/main/java/tv/trakt/trakt/core/lists/features/smart/CreateSmartListState.kt
```diff
@@ -0,0 +1,19 @@
...
+import tv.trakt.trakt.core.lists.model.SmartListItem
+
+@Immutable
+internal data class CreateSmartListState(
+    val items: ImmutableList<SmartListItem>? = null,
+    val collection: UserCollectionState = UserCollectionState.Default,
+    val filters: SmartListFilters = SmartListFilters.Default,
+    val creating: LoadingState = LoadingState.Idle,
+    val loading: LoadingState = LoadingState.Idle,
+    val error: Exception? = null,
```
michaldrabik, line 17: medium

The screen state carries a raw `Exception`, and `CreateSmartListScreen` surfaces `state.error?.message` directly in the snackbar. Ktor and serialization failures carry technical, non-localised messages (for example `Client request invalid: 500 Internal Server Error`), so the user can end up seeing raw internal text. To keep raw exceptions out of the UI, we should map failures to a typed error at the boundary and hold a localised message (or a typed error the screen resolves to an `error_*` string) in the state.

```kotlin
val error: SmartListError? = null,
```

michaldrabik (commented):
## Code Review

This pull request implements smart list creation on phone, adding a `CreateSmartList` screen, view model, state, navigation, two filter panels (simple and advanced), a bottom-sheet composer, and `CreateSmartListUseCase` / `GetSmartListPreviewUseCase`. It also reworks the `SmartList` / `SmartListFilters` domain models to carry the full filter set and wires the create/preview endpoints through `UserSmartListsRemoteDataSource`.

Feedback on the changes focuses on a mapping path that can crash while parsing smart lists from the API, and on surfacing raw exception text to the user instead of a typed, localised error.

PR #327, opened 2026-08-05
feat: all streaming details

michaldrabik:


app/src/main/java/tv/trakt/trakt/core/streamings/ui/AllStreamingsSourceRow.kt
```diff
@@ -0,0 +1,387 @@
...
+            textAlign = TextAlign.Center,
+            overflow = TextOverflow.Ellipsis,
+        )
+
+        if (!price.isNullOrBlank()) {
+            Text(
+                text = price,
+                color = TraktTheme.colors.textSecondary,
+                style = TraktTheme.typography.meta.copy(
+                    fontSize = 10.sp,
```
michaldrabik, line 232: high

The price label styles its text with an inline `fontSize = 10.sp` literal via `meta.copy(...)`. Inline `TextStyle` font-size literals are discouraged because they bypass the design system and do not scale with the typography tokens. We should add a semantic style to `TraktTheme.typography` and use it here instead.

```kotlin
style = TraktTheme.typography.metaSmall,
```

michaldrabik (commented):
## Code Review

This pull request consolidates the streaming ("Where to Watch") domain into `:common`, replacing the three per-media `GetStreamingsUseCase` variants (and their phone/TV duplicates) with a single `StreamingApiClient` plus `GetStreamingsUseCase`, `GetAllStreamingsUseCase`, and `GetPriorityStreamingUseCase` driven by a typed `StreamingsRequest`. It also adds a phone "All streamings" screen with its own ViewModel, navigation, skeleton, and source-row UI, and rewires the movie/show/episode detail flows to the shared use cases.

The consolidation reads well and removes a lot of duplication, including the old top-level `popularCountries` list that was mutated on every call. Feedback below focuses on a couple of theming and state-handling details in the new phone UI: an inline font-size literal that should become a typography token, an error state that is never surfaced to the user, a non-exhaustive `when` over a domain enum, and some magic-number paddings.

app/src/main/java/tv/trakt/trakt/core/streamings/AllStreamingsScreen.kt
```diff
@@ -0,0 +1,322 @@
...
+                )
+            }
+            items(count = 1, key = { "skeleton-$it" }) {
+                AllStreamingsSkeletonRow(
+                    modifier = Modifier.padding(start = horizontalPadding),
+                )
+            }
+        }
+
+        if (state.loading == Done && state.sections.isEmpty()) {
```
michaldrabik: medium

`AllStreamingsState` carries an `error` field, but `ContentList` only branches on `loading` and `sections`. When `loadData()` fails it still lands in the `finally` block and sets `loading = Done` with empty `sections`, so a network error renders the "No services" empty view rather than an error. We should distinguish a genuine error from an empty result so the user isn't told there are no services when the request actually failed.

```kotlin
if (state.loading == Done && state.error != null) {
    item(key = "error") {
        ContentErrorView(modifier = Modifier.padding(horizontal = horizontalPadding))
    }
} else if (state.loading == Done && state.sections.isEmpty()) {
    item(key = "empty") {
        ContentEmptyView(modifier = Modifier.padding(horizontal = horizontalPadding))
    }
}
```

app/src/main/java/tv/trakt/trakt/core/streamings/ui/AllStreamingsSourceRow.kt
```diff
@@ -0,0 +1,387 @@
...
+        .memoryCacheKey(url)
+        .placeholderMemoryCacheKey(url)
+        .build()
+}
+
+private fun StreamingService.priceLabel(type: StreamingType): String? {
+    val amount = when (type) {
+        PURCHASE -> purchasePrice
+        RENT -> rentPrice
+        else -> null
```
michaldrabik: medium

`priceLabel` matches on the `StreamingType` enum with an `else -> null` branch. If a new streaming type is added later, it will silently fall through to `null` here instead of failing the build. We should make the `when` exhaustive so new variants are caught at compile time.

```kotlin
val amount = when (type) {
    PURCHASE -> purchasePrice
    RENT -> rentPrice
    SUBSCRIPTION, FREE, FAVORITE -> null
}
```

app/src/main/java/tv/trakt/trakt/core/streamings/ui/AllStreamingsSourceRow.kt
```diff
@@ -0,0 +1,387 @@
...
+) {
+    val context = LocalContext.current
+    val contentColor = TraktTheme.colors.textPrimary
+
+    Box(
+        contentAlignment = Alignment.Center,
+        modifier = modifier
+            .size(TileSize)
+            .border(1.dp, TraktTheme.colors.separator, TileShape)
+            .padding(horizontal = 12.dp),
```
michaldrabik, line 138: low

The source and country tiles use magic-number paddings and spacing (`padding(horizontal = 12.dp)` here, `padding(horizontal = 8.dp)` and `spacedBy(3.dp)` in `ServiceCountryTile`). Where a value maps to the design system, we should reach for a `TraktTheme.spacing` token so spacing stays consistent across form factors.

```kotlin
.padding(horizontal = TraktTheme.spacing.sm)
```

PR #330, opened 2026-08-06
feat: introduce all ratings feature/sheet

michaldrabik:
Preview:

image

image


app/src/main/java/tv/trakt/trakt/core/summary/ui/header/DetailsHeader.kt
```diff
@@ -267,14 +280,30 @@ internal fun DetailsHeader(
...
+                var allRatingsVisible by remember { mutableStateOf(false) }
                 DetailsRatings(
                     traktRatings = traktRatings,
                     externalRatings = externalRatings,
                     rottenEnabled = externalRottenVisible,
+                    malEnabled = externalMalVisible,
+                    modifier = Modifier
+                        .padding(top = 16.dp)
+                        .onClick {
+                            allRatingsVisible = true
```
michaldrabik, line 292: high

The ratings row sets `allRatingsVisible = true` unconditionally, but `AllRatingsSheet` only renders when `ratings != null`. If the user taps the row while external ratings are still loading (or the request failed), nothing happens visually — yet the flag stays `true`, so the sheet will pop open on its own the moment `externalRatings` arrives. To prevent the sheet appearing unprompted, we should only arm the click when there is something to show.

```kotlin
modifier = Modifier
    .padding(top = 16.dp)
    .onClick(enabled = externalRatings != null) {
        allRatingsVisible = true
    },
```

app/src/main/java/tv/trakt/trakt/core/ratings/allratings/ui/QualityOverTimeCard.kt
```diff
@@ -0,0 +1,372 @@
...
+    seasons: ImmutableList<Season>,
+    modifier: Modifier = Modifier,
+) {
+    val dotLabelColor = TraktTheme.colors.textPrimary
+    val dotLabelStyle = TraktTheme.typography.cardTitle.copy(fontSize = 10.sp)
+    val textMeasurer = rememberTextMeasurer()
+
+    Canvas(modifier = modifier) {
+        val ratings = seasons.map { it.rating.ratingPercent }
+        val minRating = ratings.min()
```
michaldrabik: medium

`QualityLineGraph` assumes at least two seasons: `ratings.min()` throws `NoSuchElementException` on an empty list, and `size.width / (ratings.size - 1)` divides by zero for a single season, producing `Infinity`/`NaN` coordinates. The only current call site guards with `size >= 2`, but the card is `internal` and previewable, so the invariant lives far from the code that depends on it. Following the early-exit guard convention, we can make the canvas self-protecting.

```kotlin
Canvas(modifier = modifier) {
    if (seasons.size < 2) return@Canvas

    val ratings = seasons.map { it.rating.ratingPercent }
    ...
}
```

app/src/main/java/tv/trakt/trakt/core/ratings/allratings/AllRatingsState.kt
```diff
@@ -0,0 +1,12 @@
+package tv.trakt.trakt.core.ratings.allratings
+
+import androidx.compose.runtime.Immutable
+import kotlinx.collections.immutable.ImmutableList
+import tv.trakt.trakt.common.helpers.LoadingState
+import tv.trakt.trakt.common.model.Season
+
+@Immutable
+internal data class AllRatingsState(
```
michaldrabik, line 9: medium

The screen state is modelled as a flat `data class` with a nullable `seasons` plus a separate `LoadingState` field, which allows contradictory combinations (e.g. `loading = Done` with `seasons = null` vs `loading = Idle` after an error — both render the same). The architecture convention for new screens is a sealed `UiState` with `Loading` / `Loaded` / `Error` variants, which makes the `when` in the view exhaustive and removes the null-plus-flag coupling. We should consider modelling it as a sealed hierarchy.

```kotlin
@Immutable
internal sealed interface AllRatingsState {
    data object Loading : AllRatingsState
    data class Loaded(val seasons: ImmutableList<Season>) : AllRatingsState
    data object Error : AllRatingsState
}
```

app/src/main/java/tv/trakt/trakt/core/summary/ui/DetailsRatings.kt
```diff
@@ -48,209 +52,215 @@ internal fun DetailsRatings(
...
+                        ),
+                        hasRating = hasRating,
+                        valueText = "$rottenRating%",
+                        emptyText = emptyText,
+                        placeholder = "00%",
+                        textStyle = textStyle,
+                        grayFilter = grayFilter,
+                        iconSpace = iconSpace,
+                        imageModifier = Modifier.size(14.dp),
+//                    },
```
michaldrabik: low

There is a leftover fragment of commented-out code from the refactor — a stray `//                    },` line inside the Rotten Tomatoes item. We should remove it to keep the call site clean.

```kotlin
                        iconSpace = iconSpace,
                        imageModifier = Modifier.size(14.dp),
                    )
```

michaldrabik (commented):
## Code Review

This pull request introduces the new All Ratings feature: a bottom sheet opened from the details header ratings row that surfaces the Trakt rating with a 5-star vote distribution chart, a per-season "Quality Over Time" line graph (with a loading skeleton and horizontal scrolling for long-running shows), and tiles for external ratings (IMDb, MyAnimeList, Rotten Tomatoes critic/audience, TMDB). It extends `ExternalRating` with a `TraktRating` model mapped from the ratings endpoints, wires TMDB/MAL ratings through the phone and TV use cases, and adds a parametrised `AllRatingsViewModel` that fetches season ratings for shows.

Feedback on the changes focuses on a stale sheet-visibility flag that can make the sheet appear unprompted when the ratings row is tapped before external ratings have loaded, defensive guards missing in the new line-graph canvas for degenerate season lists, the new sheet state shape diverging from the sealed `UiState` convention, and a leftover commented-out fragment in `DetailsRatings`.

PR #338, opened 2026-08-14
feat: widgets

michaldrabik:
- this PR introduces 3 initial core widgets: Continue Watching, Calendar (current week) and Streaks
- dynamic sizing is possible
- widgets refresh every 30min or when certain main app events are triggered

image

image

image


app/src/main/java/tv/trakt/trakt/core/main/MainScreen.kt
```diff
@@ -673,6 +683,43 @@ private fun handleShortcutIntent(
...
+    }
+
+    val targetJson = intent.getStringExtra(INTENT_WIDGET_TARGET_EXTRA)
+    if (targetJson.isNullOrBlank()) {
+        return
+    }
+
+    intent.removeExtra(INTENT_WIDGET_TARGET_EXTRA)
+
+    when (val target = Json.decodeFromString<WidgetIntentTarget>(targetJson)) {
```
michaldrabik: high

When a widget is placed, its click PendingIntents — including the serialized `WidgetIntentTarget` extra — persist in the launcher until the widget next re-renders. If the app updates and the `WidgetIntentTarget` schema changes (a renamed subclass or field), `Json.decodeFromString` throws a `SerializationException` inside the `LaunchedEffect`, crashing the app the moment the user taps a stale widget. The same applies to any malformed extra arriving on this exported activity. To keep a bad payload from taking down the whole screen, we should decode defensively and ignore targets we can't parse.

```kotlin
    val target = runCatching {
        Json.decodeFromString<WidgetIntentTarget>(targetJson)
    }.getOrElse { error ->
        Timber.w(error, "Failed to decode the widget intent target")
        return
    }

    when (target) {
```

app/src/main/java/tv/trakt/trakt/core/calendar/usecases/GetCalendarItemsUseCase.kt
```diff
@@ -37,6 +42,7 @@ internal class GetCalendarItemsUseCase(
         day: LocalDate,
         filters: GlobalFilter,
         type: ReleaseType,
+        skipProgress: Boolean = false,
```
michaldrabik, line 45: medium

The new `skipProgress` parameter is never set to `true` by any call site — `CalendarWidgetDataSource` calls `getCalendarItems(day, filters, type)` with the default, so both `if (skipProgress)` branches are dead code. Wiring it into the widget would also break the cards' `watched` state, which is derived from the progress loaded here (lines 133 and 161). If skipping the progress load has no remaining caller, we should remove the parameter and the two branches; if the widget was meant to use it, the flag needs to be passed from `CalendarWidgetDataSource` and the widget's `watched` handling adjusted accordingly.

```kotlin
    suspend fun getCalendarItems(
        day: LocalDate,
        filters: GlobalFilter,
        type: ReleaseType,
    ): ImmutableMap<LocalDate, ImmutableList<CalendarItem>> {
```

michaldrabik (commented):
## Code Review

This pull request introduces the first three home-screen widgets — Continue Watching, Calendar (current week) and Streaks — built on Glance, together with a shared widget configuration activity (background style + header visibility), per-widget appearance stored in Glance state, deep-link intents into the main NavHost, and refresh triggers wired into the history/watchlist sync use cases so widgets update on relevant app events as well as on the 30-minute platform cycle. It also generalises `UpNextUpdates` into a source-scoped update bus and retires the duplicate `HomeUpNextLocalDataSource` notification channel.

Feedback on the changes focuses on the robustness of the widget deep-link decoding (a stale serialized target can currently crash the app on tap), a duplicated data refresh on the widget's mark-as-watched path, an unused `skipProgress` parameter, and a few localization gaps in the widget-facing strings ("N/A" literals and a hand-assembled streak label that won't survive word-order differences across locales).

app/src/main/java/tv/trakt/trakt/widgets/continuewatching/actions/MarkWatchedAction.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        val widget = ContinueWatchingWidget(dataSource)
+
+        dataSource.setPendingItem(itemKey)
+        // A live session recomposes off the data source; this covers an idle widget without one.
+        widget.update(context, glanceId)
+
+        try {
+            addToHistoryUseCase.addToHistory(episodeId = episodeId.toTraktId())
+            dataSource.refresh(context = context, limit = MAX_ITEM_COUNT)
```
michaldrabik: medium

`addToHistoryUseCase.addToHistory(...)` goes through `UpdateEpisodeHistoryUseCase`, which now calls `widgetsUpdater.refreshInBackground()` on success — that already refreshes the Continue Watching data source (plus Calendar and Streaks). The explicit `dataSource.refresh(...)` on the next line then fetches the same up-next data a second time, so a single tap costs two concurrent up-next requests and two rounds of bitmap decodes for the same result. Given `ContinueWatchingWidgetUpdater`'s own doc puts the "did anything change" cost question on the caller, we should avoid the double fetch — for example by letting the widget path skip the use case's blanket widget refresh, since this action already refreshes and repaints the widget itself.

```kotlin
// UpdateEpisodeHistoryUseCase
suspend fun addToHistory(
    episodeId: TraktId,
    watchedAt: DateSelectionResult? = null,
    notifyWidgets: Boolean = true,
) {
    // ...
    if (notifyWidgets) widgetsUpdater.refreshInBackground()
}

// WidgetAddToHistoryUseCase
updateHistoryUseCase.addToHistory(episodeId = episodeId, notifyWidgets = false)
```

app/src/main/java/tv/trakt/trakt/widgets/continuewatching/data/ContinueWatchingWidgetDataSource.kt
```diff
@@ -0,0 +1,187 @@
...
+            .build()
+
+        return when (val result = imageLoader.execute(request)) {
+            is SuccessResult -> result.image.toBitmap()
+            else -> null
+        }
+    }
+
+    private companion object {
+        const val UNKNOWN_TEXT = "N/A"
```
michaldrabik: medium

`UNKNOWN_TEXT = "N/A"` (and the second literal in `durationText()` at line 187) is user-facing copy rendered on the widget cards, but it's an inline raw string, so it will show as English "N/A" in every locale. User-facing text belongs in `resources/.../values/strings.xml`; we should add a string resource and resolve it through the `context` that's already threaded through these mappers.

```kotlin
episodeText = progress.nextEpisode?.seasonEpisodeText(context)
    ?: context.getString(R.string.text_value_unknown),
```

app/src/main/java/tv/trakt/trakt/widgets/streaks/StreaksWidget.kt
```diff
@@ -0,0 +1,363 @@
...
+}
+
+@Composable
+private fun StreakLabelView(
+    streakDays: Int,
+    isShort: Boolean,
+) {
+    val context = LocalContext.current
+
+    val fullText = "${
```
michaldrabik, line 196: medium

The streak label is assembled by branching on `streakDays > 1` and interpolating two separate resources (`text_stats_days_count` + `text_stats_watching_streak`) with a hard-coded space / newline. Pluralisation via an `if count == 1` style branch breaks locales with more than two plural forms, and joining two independently translated fragments fixes the word order, which many languages need to reorder. We should express this as a single `<plurals>` entry that contains the whole phrase, and resolve it once.

```xml
<plurals name="text_widget_watching_streak">
    <item quantity="one">%d-day streak</item>
    <item quantity="other">%d-day streak</item>
</plurals>
```

```kotlin
val text = context.resources.getQuantityString(
    R.plurals.text_widget_watching_streak,
    streakDays,
    streakDays,
)
```

app/src/main/java/tv/trakt/trakt/widgets/streaks/StreaksWidgetState.kt
```diff
@@ -0,0 +1,17 @@
+package tv.trakt.trakt.widgets.streaks
+
+import kotlinx.collections.immutable.ImmutableList
+import kotlinx.collections.immutable.persistentListOf
+
+internal data class StreaksWidgetState(
```
michaldrabik, line 8: low

`CalendarWidgetState` and `ContinueWatchingWidgetState` are annotated `@Immutable`, but `StreaksWidgetState` and `StreaksWidgetDay` are not, even though they're held in the Glance composition state the same way. For consistency and to keep the stability contract explicit, we should add the annotation here too.

```kotlin
@Immutable
internal data class StreaksWidgetState(
```

app/src/main/java/tv/trakt/trakt/widgets/streaks/StreaksWidgetUpdater.kt
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    fun refreshInBackground() {
+        scope.launch {
+            delay(3.seconds)
+            refresh()
+        }
+    }
+
```
michaldrabik, line 40: low

Unlike the other two updaters, `refreshInBackground` here waits a fixed 3 seconds before refreshing, and nothing explains why. Every caller pays it, including the app-start refresh from `MainViewModel.loadData()`. If the delay exists to let the server register a just-posted play before the streak is recomputed, we should say so in a comment (and ideally scope it to the history-add trigger rather than every refresh); otherwise it can be removed.

```kotlin
    fun refreshInBackground() {
        scope.launch {
            // Give the server a moment to register a just-posted play before
            // the streak is recomputed.
            delay(3.seconds)
            refresh()
        }
    }
```

PR #342, opened 2026-08-18
feature: light theme

michaldrabik:


app/src/main/AndroidManifest.xml
```diff
@@ -125,7 +125,7 @@
     </activity>
 
     <receiver
-      android:name=".widgets.continuewatching.ContinueWatchingWidgetReceiver"
+      android:name=".widgets.widget.continuewatching.ContinueWatchingWidgetReceiver"
```
michaldrabik, line 128: critical

The three widget receivers moved from `tv.trakt.trakt.widgets.<feature>` to `tv.trakt.trakt.widgets.widget.<feature>`, and the manifest was updated to match. The launcher, however, persists the fully-qualified `ComponentName` of the provider for every widget a user has already placed. After this update those components no longer exist, so existing Calendar, Continue Watching and Streaks widgets will stop updating and render as blank or removed tiles, and the per-widget Glance state keyed to the old provider is orphaned along with them. Users would have to delete and re-add every widget. To keep placed widgets alive, we should either revert the package move for the three receiver classes, or leave a thin subclass behind at each original FQN and keep the old `<receiver>` entries in the manifest.

```kotlin
// app/src/main/java/tv/trakt/trakt/widgets/calendar/CalendarWidgetReceiver.kt
package tv.trakt.trakt.widgets.calendar

/** Kept at the original FQN: launchers store this ComponentName for already-placed widgets. */
internal class CalendarWidgetReceiver : tv.trakt.trakt.widgets.widget.calendar.CalendarWidgetReceiver()
```

app/src/main/java/tv/trakt/trakt/MainActivity.kt
```diff
@@ -80,8 +92,41 @@ internal class MainActivity : AppCompatActivity() {
             ),
         )
 
+        val initialThemeMode = runBlocking {
```
michaldrabik: high

`runBlocking { themeModeUseCase.getThemeMode() }` blocks the main thread inside `onCreate`, before `setContent` runs, while DataStore reads and deserialises its file from disk. This is on the cold-start critical path, so a slow or contended disk turns directly into startup jank or an ANR, and the style guide lists `runBlocking` outside tests as block-worthy. Since `observeThemeMode()` is already collected below, we can seed the collection with `ThemeMode.Default` and drop the blocking read. If avoiding a first-frame flash is the reason for reading synchronously, gating on `SplashScreen.setKeepOnScreenCondition` until the first value arrives gets the same result without blocking.

```kotlin
val themeMode by themeModeUseCase.observeThemeMode()
    .collectAsStateWithLifecycle(initialValue = ThemeMode.Default)
```

app/src/main/java/tv/trakt/trakt/core/streamings/ui/JustWatchRanksStrip.kt
```diff
@@ -47,16 +50,20 @@ internal fun JustWatchRanksStrip(
             )
         },
     ) {
+        val jwColor = when (TraktTheme.colors.isLight) {
```
michaldrabik: high

The JustWatch accent is picked with raw hex literals switched on `TraktTheme.colors.isLight`, then wrapped at three call sites via `Color(jwColor)`. Raw `Color(0xFF…)` in feature code is an anti-pattern in the style guide, and the fact that the value now differs per theme is exactly the signal that it belongs in the palette rather than in the composable - as written, a future seasonal or contrast tweak has to be made here instead of in `TraktColors`. We should add a token with light and dark variants and read it through `TraktTheme.colors`.

```kotlin
// TraktColors.kt
val justWatchAccent: Color = Color.Unspecified,

// ColorLight.kt -> justWatchAccent = JustWatchYellowDim,
// ColorDark.kt  -> justWatchAccent = JustWatchYellow,

// JustWatchRanksStrip.kt
Icon(
    painter = painterResource(R.drawable.ic_justwatch),
    contentDescription = null,
    tint = TraktTheme.colors.justWatchAccent,
    modifier = Modifier.size(14.dp),
)
```

app/src/main/java/tv/trakt/trakt/core/settings/SettingsScreen.kt
```diff
@@ -953,27 +1004,31 @@ private fun RateTraktView(modifier: Modifier = Modifier) {
...
-                color = TraktTheme.colors.textPrimary,
+                color = TraktTheme.colors.textPrimaryOnAccent,
                 textAlign = TextAlign.Start,
             )
         }
     }
 }
 
+@Composable
+private fun getHeaderColor(): Color {
```
michaldrabik, line 1022: medium

`getHeaderColor()` branches on `TraktTheme.colors.isLight` and returns the `Purple500` / `Purple300` palette constants directly. The style guide asks for all colours to come from `TraktTheme.colors.*`, and a light/dark branch inside a screen is the same coupling a token exists to remove - every other settings section now calls this helper, so the choice is repeated four times in one file. We should promote it to a token so the palette owns the light and dark values and the screen just reads one name.

```kotlin
// TraktColors.kt
val sectionHeader: Color = Color.Unspecified,

// ColorLight.kt -> sectionHeader = Purple500,
// ColorDark.kt  -> sectionHeader = Purple300,

// SettingsScreen.kt
TraktHeader(
    title = stringResource(R.string.header_appearance).uppercase(),
    titleColor = TraktTheme.colors.sectionHeader,
    titleStyle = TraktTheme.typography.heading6,
)
```

michaldrabik (commented):
## Code Review

This pull request introduces a light theme across the app. It adds a `ThemeMode` preference (System / Light / Dark) backed by a new `ThemeModeUseCase` over DataStore, wires the selection into `MainActivity` and the Settings screen, extends `TraktColors` with light variants plus new shadow and reaction tokens, and sweeps a large number of feature composables to read theme-aware tokens instead of dark-only constants. It also reworks widget theming so each placed widget can pick its own theme, and adds light-variant previews across many components.

Feedback on the changes focuses on a package rename that will break widgets users already have placed on their home screens, a blocking DataStore read on the cold-start path in `MainActivity`, and a few places where the new light/dark branches reach for raw colour literals or palette constants instead of going through `TraktTheme.colors`.

app/src/main/java/tv/trakt/trakt/core/summary/shows/features/seasons/all/AllShowSeasonsContent.kt
```diff
@@ -309,6 +310,53 @@ private fun PreviewLoaded() {
     }
 }
 
+@Preview(
```
michaldrabik, line 313: low

`PreviewLoadedLight` renders the light theme against `backgroundColor = 0xFF131517`, the dark surface colour copied from the neighbouring dark preview, so the preview shows light content floating on a dark page and any contrast problem it is meant to catch is masked. It also hand-rolls `TraktTheme(colors = LightColors)` while the rest of this pull request standardises on the new `TraktThemeLightDark` helper. We should use the helper here too, which renders both themes on their own backgrounds and removes the duplicated preview body.

```kotlin
@Preview(device = "id:pixel_5", showBackground = true)
@Composable
private fun PreviewLoaded() {
    TraktThemeLightDark {
        // existing preview body
    }
}
```

PR #344, opened 2026-08-19
feat(comments): introduce comments section language filter

michaldrabik:


app/src/main/java/tv/trakt/trakt/core/comments/ui/CommentsLanguageDropdown.kt
```diff
@@ -0,0 +1,192 @@
...
+                },
+        ) {
+            Icon(
+                painter = painterResource(R.drawable.ic_world),
+                contentDescription = null,
+                tint = TraktTheme.colors.textPrimary,
+                modifier = Modifier.size(18.dp),
+            )
+
+            if (language != null && language != DEFAULT_LANGUAGE) {
```
michaldrabik: high

The active indicator is suppressed for `en`, but nothing else in the feature treats English as the default: `GetCommentsLanguageUseCase.getLanguage()` returns `null` when nothing is stored, the "All languages" row is checked on `language == null`, and picking English persists `"en"` and sends `language=en` to the API. So a user who selects English gets a genuinely filtered list (English comments only) while the header claims no filter is active, and the menu simultaneously shows English as checked. This is a second, divergent definition of "default" living in the composable. We should drop the constant and derive the indicator from the same state everything else reads.

```kotlin
if (language != null) {
    Box(
        modifier = Modifier
            .align(Alignment.BottomEnd)
```

app/src/main/java/tv/trakt/trakt/core/summary/movies/features/comments/usecases/GetMovieCommentsUseCase.kt
```diff
@@ -18,6 +18,7 @@ internal class GetMovieCommentsUseCase(
         movieId: TraktId,
         user: User? = null,
         filter: CommentsFilter = Popular,
+        language: String? = null,
```
michaldrabik, line 21: medium

The new `language` parameter defaults to `null` on all three comments use-cases, and `CommentsViewModel` - the "all comments" screen reached by tapping the section header - was not updated, so it keeps calling `getComments(...)` without a language. A user who filters a movie's comments to German and then opens the full list sees every language again, with no control on that screen to explain why. We should read the stored language there too, so the persisted selection holds across both screens.

```kotlin
private suspend fun fetchComments(): ImmutableList<Comment> {
    val filter = loadFilter()
    val language = getLanguageUseCase.getLanguage()
    return when (destination.mediaType) {
        MediaType.Movie -> getMovieCommentsUseCase.getComments(
            movieId = destination.mediaId.toTraktId(),
            user = userState.value,
            filter = filter,
            language = language,
            limit = 250,
        )
```

app/src/main/java/tv/trakt/trakt/core/comments/ui/CommentsLanguageDropdown.kt
```diff
@@ -0,0 +1,192 @@
...
+                .onClick(enabled = enabled) {
+                    showMenu = true
+                }
+                .graphicsLayer {
+                    translationY = 0.75.dp.toPx()
+                },
+        ) {
+            Icon(
+                painter = painterResource(R.drawable.ic_world),
+                contentDescription = null,
```
michaldrabik, line 62: medium

The globe icon is the only affordance that opens the language filter, and it carries `contentDescription = null`, so TalkBack announces nothing for it and the filter is unreachable for screen-reader users. The comment-plus icon next to it has the same gap, but that one at least sits beside a labelled header. We should add a namespaced `a11y_*` string and pass it here, ideally reflecting the current selection so the state is announced too.

```kotlin
Icon(
    painter = painterResource(R.drawable.ic_world),
    contentDescription = stringResource(R.string.a11y_comments_language_filter),
    tint = TraktTheme.colors.textPrimary,
    modifier = Modifier.size(18.dp),
)
```

app/src/main/java/tv/trakt/trakt/core/comments/usecases/GetCommentsLanguageUseCase.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.datastore.preferences.core.stringPreferencesKey
+import kotlinx.coroutines.flow.first
+
+private val KEY_COMMENTS_LANGUAGE = stringPreferencesKey("key_comments_language")
+
+/**
+ * Comments language filter, shared by movie, show and episode comments. A null language means all
+ * languages.
+ */
+internal class GetCommentsLanguageUseCase(
```
michaldrabik, line 14: medium

Unit coverage is mandatory for use-cases and pure helpers, and this pull request adds a new use-case plus three pure helpers (`commentsLanguages()`, `apiLanguage`, `countryFlag`) with no tests. The `apiLanguage` mapping in particular is exactly the kind of logic that regresses silently: it exists because Android reports `in` for Indonesian and `iw` for Hebrew, and because Trakt files both Norwegian standards under `no`, and a wrong code returns an empty comment list rather than an error. We should cover the round-trip and the code mapping with hand-written fakes, no mocking library.

```kotlin
@Test
fun `stored language survives a round-trip and clears on null`() = runTest {
    val useCase = GetCommentsLanguageUseCase(FakePreferencesDataStore())

    useCase.setLanguage("de")
    assertThat(useCase.getLanguage()).isEqualTo("de")

    useCase.setLanguage(null)
    assertThat(useCase.getLanguage()).isNull()
}

@Test
fun `deprecated ISO codes map to the codes Trakt accepts`() {
    assertThat(Locale("in").apiLanguage).isEqualTo("id")
    assertThat(Locale("nb").apiLanguage).isEqualTo("no")
}
```

app/src/main/java/tv/trakt/trakt/core/comments/ui/CommentsLanguageDropdown.kt
```diff
@@ -0,0 +1,192 @@
...
+    onClick: () -> Unit,
+    leadingContent: @Composable (() -> Unit)? = null,
+) {
+    DropdownMenuItem(
+        text = {
+            Row(
+                horizontalArrangement = spacedBy(10.dp),
+                verticalAlignment = Alignment.CenterVertically,
+            ) {
+                if (selected) {
```
michaldrabik, line 142: low

The checkmark is rendered only when a row is selected, while the flag sits in a fixed-width `Box`. That means the selected row is wider by the check plus its spacing, and its flag and label shift right relative to every other row in the menu. We should reserve the check column the same way the flag column is reserved, so only the glyph appears and disappears.

```kotlin
Box(
    contentAlignment = Alignment.Center,
    modifier = Modifier.width(CheckSize),
) {
    if (selected) {
        Icon(
            painter = painterResource(R.drawable.ic_check),
            contentDescription = null,
            tint = TraktTheme.colors.textPrimary,
            modifier = Modifier.size(CheckSize),
        )
    }
}
```

app/src/main/java/tv/trakt/trakt/core/comments/ui/CommentsLanguageDropdown.kt
```diff
@@ -0,0 +1,192 @@
...
+                }
+                .graphicsLayer {
+                    translationY = 0.75.dp.toPx()
+                },
+        ) {
+            Icon(
+                painter = painterResource(R.drawable.ic_world),
+                contentDescription = null,
+                tint = TraktTheme.colors.textPrimary,
+                modifier = Modifier.size(18.dp),
```
michaldrabik: low

The icon and dot sizes are raw `dp` literals inline (`18.dp` here, `6.dp` for the active dot), while `FlagSize` and `CheckSize` right above them show the pattern this file already follows. Sizes should come from `TraktTheme.size.*` where a token fits; failing that, a named private `val` keeps the two call sites of the icon size in step.

```kotlin
private val WorldIconSize = 18.dp
private val ActiveDotSize = 6.dp
```

michaldrabik (commented):
## Code Review

This pull request introduces a language filter for the comments sections of movies, shows and episodes. A globe control in each section header opens a dropdown listing "All languages" followed by every language the app is localised into, the selection is persisted in the comments DataStore through a new `GetCommentsLanguageUseCase`, and the chosen code is threaded down through the use-cases and remote data sources into the `language` query parameter of the Trakt comments endpoints. It also promotes the `apiLanguage` mapping and the `countryFlag` helper into shared `common` extensions so the translations client and the new filter agree on the codes they send.

Feedback on the changes focuses on the `DEFAULT_LANGUAGE` special case in the dropdown, which makes the active indicator disagree with both the persisted selection and the request that is sent, on the "all comments" screen still fetching unfiltered while the rest of the feature honours the stored language, on the missing content description for the only control that opens the filter, and on the absence of unit coverage for the new use-case and pure helpers. Two advisory notes cover row alignment in the menu and raw `dp` sizes where theme tokens exist.

PR #350, opened 2026-08-24
feat(picker): introduce slot based other date picker

michaldrabik:
Preview:
image

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/ui/components/dateselection/otherdatepicker/OtherDatePickerView.kt
+++ b/app/src/main/java/tv/trakt/trakt/ui/components/dateselection/otherdatepicker/OtherDatePickerView.kt
@@ -1,0 +1,599 @@
+package tv.trakt.trakt.ui.components.dateselection.otherdatepicker
+
+import androidx.compose.animation.Crossfade
+import androidx.compose.animation.core.Animatable
+import androidx.compose.animation.core.keyframes
+import androidx.compose.animation.core.tween
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.layout.Arrangement.Absolute.spacedBy
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.WindowInsets
+import androidx.compose.foundation.layout.asPaddingValues
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.navigationBars
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Icon
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.derivedStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Alignment.Companion.CenterVertically
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Brush.Companion.verticalGradient
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextOverflow.Companion.Ellipsis
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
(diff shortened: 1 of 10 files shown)


app/src/main/java/tv/trakt/trakt/ui/components/dateselection/otherdatepicker/OtherDatePickerView.kt
```diff
@@ -0,0 +1,599 @@
...
+    state: OtherDatePickerState,
+    title: String,
+    modifier: Modifier = Modifier,
+    subtitle: String? = null,
+    contentPadding: PaddingValues = PaddingValues(0.dp),
+    onLoadMore: () -> Unit = {},
+    onSlotClick: (before: HomeActivityItem?, after: HomeActivityItem?) -> Unit = { _, _ -> },
+    onConfirm: (Instant) -> Unit = {},
+) {
+    var selectedDate by remember { mutableStateOf<Instant?>(null) }
```
michaldrabik, line 116: medium

If the user scrolls through their history, picks a slot, and then rotates the device (or the activity is otherwise recreated), `remember` resets `selectedDate`, `selectedSlotId`, and the slot bounds — the selection silently disappears and the confirm button turns disabled again. Since this state should survive configuration changes, we should hold it in `rememberSaveable` (storing the instant as epoch milliseconds, which is `Bundle`-friendly).

```kotlin
var selectedDateMillis by rememberSaveable { mutableStateOf<Long?>(null) }
var selectedSlotId by rememberSaveable { mutableStateOf<Long?>(null) }
val selectedDate = selectedDateMillis?.let(Instant::ofEpochMilli)
```

app/src/main/java/tv/trakt/trakt/ui/components/dateselection/otherdatepicker/OtherDatePickerView.kt
```diff
@@ -0,0 +1,599 @@
...
+        }
+    }
+
+    LaunchedEffect(isScrolledToBottom) {
+        if (isScrolledToBottom) {
+            onLoadMore()
+        }
+    }
+
+    if (state.loading.isLoading) {
```
michaldrabik, line 371: medium

`OtherDatePickerState.error` is populated by the ViewModel when the history request fails, but the content never reads it — after a failed initial load the user is left with an empty list and no way to retry, since `loadInitialData` only runs from `init`. We should render an error branch (and ideally a retry hook) before falling through to the list.

```kotlin
if (state.error != null) {
    GenericErrorView(
        onRetry = onRetry,
        modifier = modifier.fillMaxWidth(),
    )
    return
}
```

app/src/main/java/tv/trakt/trakt/ui/components/dateselection/otherdatepicker/OtherDatePickerView.kt
```diff
@@ -0,0 +1,599 @@
...
+                .onClick { onClick() },
+        ) {
+            Icon(
+                painter = painterResource(
+                    when {
+                        selected -> R.drawable.ic_check
+                        else -> R.drawable.ic_plus
+                    },
+                ),
+                contentDescription = null,
```
michaldrabik, line 478: medium

The plus/check control is the core interaction of this screen, but it is a clickable `Box` whose icon has `contentDescription = null` — TalkBack users cannot discover or activate the insert slots. We should describe the action on the clickable container (an `a11y_*` string), and reflect the selected state.

```kotlin
.onClick { onClick() }
.semantics {
    contentDescription = context.getString(R.string.a11y_insert_history_slot)
},
```

app/src/main/java/tv/trakt/trakt/ui/components/dateselection/otherdatepicker/OtherDatePickerView.kt
```diff
@@ -0,0 +1,599 @@
...
+                    is MovieItem -> {
+                        item.movie.title
+                    }
+                    is EpisodeItem -> {
+                        val episodeTitle = stringResource(
+                            R.string.episode_footer_season_episode,
+                            item.episode.season,
+                            item.episode.number,
+                        )
+                        "$episodeTitle - ${item.show.title}"
```
michaldrabik, line 528: medium

`"$episodeTitle - ${item.show.title}"` concatenates user-facing text in code, which locks the " - " separator and the ordering for every locale. Since translators can't reorder or adapt this, we should compose the line through a single placeholder template in `strings.xml` instead.

```kotlin
is EpisodeItem -> stringResource(
    R.string.text_history_item_episode_title,
    stringResource(
        R.string.episode_footer_season_episode,
        item.episode.season,
        item.episode.number,
    ),
    item.show.title,
)
```

```xml
<string name="text_history_item_episode_title">%1$s - %2$s</string>
```

app/src/main/java/tv/trakt/trakt/ui/components/dateselection/otherdatepicker/OtherDatePickerView.kt
```diff
@@ -0,0 +1,599 @@
...
+                    }
+                },
+                color = TraktTheme.colors.textPrimary,
+                style = TraktTheme.typography.cardTitle.copy(fontSize = 16.sp),
```
michaldrabik, line 532: medium

Both row texts override the theme's type scale inline — `cardTitle.copy(fontSize = 16.sp)` here and `cardSubtitle.copy(fontSize = 12.sp)` below. Inline font-size literals bypass the semantic typography tokens and won't track Dynamic Type or future scale changes. We should pick the `TraktTheme.typography` style that already carries the intended size (or add one) rather than patching sizes at the call site.

```kotlin
style = TraktTheme.typography.cardTitle,
```

michaldrabik (commented):
## Code Review

This pull request introduces a slot-based "other date" picker for marking items as watched. A new `OtherDatePicker` sheet loads the user's paginated watch history, lets them insert the new watch between two neighbouring history entries (defaulting to the midpoint timestamp), and fine-tune the result with range-constrained date and time pickers. It also wires the new sheet into `DateSelectionSheet` in place of the previous bare date/time picker flow, adds a `contentWindowInsets` pass-through on `TraktBottomSheet`, a loading skeleton, and a reusable `RangeSelectableDates`.

Feedback on the changes focuses on state management and resilience in the new view — preserving the selected slot across configuration changes and surfacing the load-failure state instead of an empty list — along with a few style-guide items: localisation-safe title composition, accessibility of the tappable slot controls, and typography tokens without inline font-size overrides.

PR #351, opened 2026-08-25
feat: all streamings screen UI update

michaldrabik:


app/src/main/java/tv/trakt/trakt/core/streamings/AllStreamingsScreen.kt
```diff
@@ -167,17 +216,47 @@ private fun ContentList(
...
                 key = { "${section.type.type}-${it.source}" },
             ) { row ->
-                AllStreamingsSourceRow(
+                val groupKey = "${section.type.type}-${row.source}"
+                AllStreamingsServiceGroup(
                     row = row,
                     type = section.type,
-                    startPadding = horizontalPadding,
-                    endPadding = horizontalPadding,
+                    expanded = query.isNotBlank() || groupKey in expandedGroups,
```
michaldrabik, line 223: medium

While a search query is active, `expanded` is forced to `true`, but `onToggleExpand` still mutates `expandedGroups`. Tapping the chevron during a search appears to do nothing (the group stays expanded), yet it silently adds or removes keys from the saved set — so after clearing the query, groups the user tapped mid-search end up in a surprising expand state. We should ignore toggle events while the query is active so the persisted state only changes when the user can actually see the effect.

```kotlin
expanded = query.isNotBlank() || groupKey in expandedGroups,
onToggleExpand = {
    if (query.isBlank()) {
        expandedGroups = when (groupKey) {
            in expandedGroups -> expandedGroups - groupKey
            else -> expandedGroups + groupKey
        }
    }
},
```

app/src/main/java/tv/trakt/trakt/core/streamings/ui/AllStreamingsServiceGroup.kt
```diff
@@ -0,0 +1,507 @@
...
+            horizontalArrangement = spacedBy(12.dp),
+            modifier = Modifier
+                .weight(1F)
+                .padding(end = 24.dp),
+        ) {
+            SourceLogoTile(
+                service = source,
+                type = type,
+                modifier = Modifier
+                    .onClick(onClick = onCountryClick),
```
michaldrabik: medium

The logo tile and the name/subtitle column both carry `onCountryClick`, which opens an external app/link, while the rest of the header row toggles expansion. In the flags-only variant (2–6 countries) the header shows several flags with no indication of which country a tap would open, and the logo tile — the most prominent tap target — launches the first country's link instead of expanding the group. Since every country is already reachable through the expanded grid in this mode, we should make the whole header toggle-only when `flagsOnly` is true and keep the direct link tap for the single-country / name variant.

```kotlin
SourceLogoTile(
    service = source,
    type = type,
    modifier = Modifier
        .onClick(onClick = if (flagsOnly) onClick else onCountryClick),
)

Column(
    verticalArrangement = spacedBy(4.dp),
    modifier = Modifier
        .onClick(onClick = if (flagsOnly) onClick else onCountryClick),
) {
```

app/src/main/java/tv/trakt/trakt/core/streamings/ui/AllStreamingsServiceGroup.kt
```diff
@@ -0,0 +1,507 @@
...
+                color = TraktTheme.colors.textPrimary,
+                style = TraktTheme.typography.cardTitle,
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis,
+            )
+            if (!price.isNullOrBlank()) {
+                Text(
+                    text = price,
+                    color = TraktTheme.colors.textSecondary,
+                    style = TraktTheme.typography.cardSubtitle.copy(
```
michaldrabik, line 406: low

The price label builds a hybrid style with `cardSubtitle.copy(fontSize = meta.fontSize)`. This sidesteps the semantic typography tokens — if `cardSubtitle` or `meta` changes, this composite drifts in a way no token owner can see. We can use `TraktTheme.typography.meta` directly (it is already used for the `+N` count in the header), or add a dedicated token if this exact combination is intentional.

```kotlin
Text(
    text = price,
    color = TraktTheme.colors.textSecondary,
    style = TraktTheme.typography.meta,
    maxLines = 1,
    overflow = TextOverflow.Ellipsis,
)
```

michaldrabik (commented):
## Code Review

This pull request revamps the "Where to Watch" (All Streamings) screen with a new vertically stacked, collapsible group layout. Each streaming service now renders as a header card (logo tile, service name, primary country or flag strip) with an expandable two-column grid of country chips, replacing the old horizontal tile rows. It also adds a local search field filtering by service name or country, a search-bar skeleton, tinted logos reusing a newly shared `getContrastRatio` helper in `:common`, reordered `StreamingType` sections, and drops the user-country restriction for free/purchase/rent listings.

Feedback on the changes focuses on the interaction between the search query and the manual expand/collapse state, the ambiguity of the tap targets inside the group header in the flags-only variant, and a small typography token workaround in the country chip.

PR #354, opened 2026-08-26
feat(app): split up to date progress filter into up to date and ended

### LINKED ISSUE #353: Split the "Up to date" progress filter into up to date and ended, with counts on each filter
### Code of Conduct

- [x] I have read and agree to the project's [Code of Conduct](https://github.com/trakt/trakt-web/blob/main/.github/CODE_OF_CONDUCT.md)

The "Up to date" filter in Profile > Progress answers two different questions at once. It lists every show whose aired episodes I have all watched, so a series that wrapped up years ago sits right next to a series that is still airing and simply has nothing unwatched at the moment.

That difference matters a lot to me as a user. "Better Call Saul is done, there is nothing left to watch" and "The Pitt is caught up, a new episode is coming" are very different states, and today the app gives me no way to tell them apart without opening each show one by one.

Split "Up to date" into two filters:

- **Up to date**: still airing (status is anything other than ended or canceled), and I am caught up.
- **Ended**: status is ended or canceled, and fully watched. Nothing more is coming.

And show how many shows sit behind each filter in parentheses, on all four chips: `Up to date (6)`, `Ended (10)`, `In progress (14)`, `Dropped (0)`.

The up next response already carries `show.status`, so this needs no API change. On my own library the split comes out as 6 still airing against 10 finished, which today are a single undifferentiated list of 16.

- Keeping a single filter and grouping the list into two sections with headers. Less discoverable, and it does not give you the number next to the chip.
- A sort order that pushes ended shows to the top. Still one mixed list.
- Filtering server side. The up next endpoint has no status filter, so that would mean an API change; doing the split on the client avoids it.

### Additional context

I have a working implementation on a branch and I am happy to open a PR if you are interested in the feature.

Two notes on the approach, in case you would rather see it done differently:

- Exact counts need each bucket loaded in full, and since the endpoint cannot filter by status the partition has to happen on device too. My branch loads the buckets to exhaustion behind a page cap, which means the all progress list arrives complete rather than growing as you scroll.
- No new translatable string is needed. "Up to date" keeps `button_text_progress_completed` and "Ended" reuses `translated_value_status_ended`, both already translated in every locale, so the feature ships localised straight away.

Screenshots to follow.

matheus-souza:
Closes #353

Scoped to the split alone: no counts, no settings toggle, no API change. Mirrors trakt/trakt-web#3211, which does the same thing on the web.

### Why

The `Up to date` filter in Profile > Progress answers two different questions at once. It returns every show whose aired episodes are all watched, so a series that wrapped up years ago sits next to a series that is still airing and simply has nothing unwatched right now. From the Progress section there is no way to tell which shows are actually finished without opening each one.

### What changed

`Up to date` becomes two filters: **Up to date** (still airing, caught up) and **Ended** (status `ended` or `canceled`). `In progress` and `Dropped` are untouched.

- **No API change.** `up_next_nitro` returns the full show object with `status`, so both filters read the same `intent=completed` bucket and partition it on that status. Verified against the live API: 16 completed, 10 `ended`, 6 `returning series`, 0 without a status.
- **`Show.hasEnded` sits in `:common`** next to `Show.isReleased`, since it is a question about a show rather than about this screen. An unknown status counts as still airing, which preserves today's behaviour.
- **Paging keys off the raw page size.** This is the one subtlety. Partitioning a page client-side shrinks it, and the pager previously decided `hasMoreData` from the size of what it received. Left alone, a page of 100 completed shows that filters down to 8 ended ones would have ended pagination early. `ProgressPage` carries the count the endpoint actually returned alongside the filtered items, and the pager reads that.
- **The chips need no UI change.** `ProgressFilters` already iterates `ProgressFilter.entries`, so adding the enum entry is enough. `ic_flag_checker` already existed.
- **The `ProgressFilter` entry `Completed` became `UpToDate`,** so a stored `"Completed"` no longer parses. `GetProgressFilterUseCase` already falls back to the default on an unknown value, and the default is `UpToDate`, which carries the same label users saw before. No migration needed.

### Localisation

No new translatable keys. `Up to date` keeps `button_text_progress_completed` and `Ended` reuses `translated_value_status_ended`, both already translated in every locale, so this ships localised rather than waiting on a Crowdin round trip.

### Notes

One consequence of partitioning client-side: the profile Progress row asks for `PROGRESS_SECTION_LIMIT` items and then splits them, so that preview row can show fewer than the limit. The all-progress screen is unaffected, since it pages. Calling it out rather than hiding it; the alternative would be over-fetching to fill the row, which is not worth it for a preview.

### Testing

`./gradlew :app:compileInternalDebugKotlin` clean, ktlint 1.7.1 (the version CI runs) clean.

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/core/profile/sections/progress/ProfileProgressView.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/profile/sections/progress/ProfileProgressView.kt
@@ -275,7 +275,7 @@ private fun Preview3() {
         ProfileProgressContent(
             state = ProfileProgressState(
                 loading = Done,
-                filter = ProgressFilter.Completed,
+                filter = ProgressFilter.UpToDate,
                 items = EmptyImmutableList,
             ),
         )
--- a/app/src/main/java/tv/trakt/trakt/core/profile/sections/progress/ProfileProgressViewModel.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/profile/sections/progress/ProfileProgressViewModel.kt
@@ -31,9 +31,11 @@ import tv.trakt.trakt.core.lists.ListsConfig.PROGRESS_SECTION_LIMIT
 import tv.trakt.trakt.core.profile.sections.progress.filters.GetProgressFilterUseCase
 import tv.trakt.trakt.core.profile.sections.progress.model.ProfileProgressItem
 import tv.trakt.trakt.core.profile.sections.progress.model.ProgressFilter
-import tv.trakt.trakt.core.profile.sections.progress.model.ProgressFilter.Completed
 import tv.trakt.trakt.core.profile.sections.progress.model.ProgressFilter.Dropped
+import tv.trakt.trakt.core.profile.sections.progress.model.ProgressFilter.Ended
 import tv.trakt.trakt.core.profile.sections.progress.model.ProgressFilter.InProgress
+import tv.trakt.trakt.core.profile.sections.progress.model.ProgressFilter.UpToDate
+import tv.trakt.trakt.core.profile.sections.progress.model.pageOfEnded
 import tv.trakt.trakt.core.profile.sections.progress.usecase.GetProgressCompleteUseCase
 import tv.trakt.trakt.core.profile.sections.progress.usecase.GetProgressDroppedUseCase
 import tv.trakt.trakt.core.profile.sections.progress.usecase.GetProgressWatchingUseCase
@@ -103,7 +105,8 @@ internal class ProfileProgressViewModel(
 
                 itemsState.update {
                     when (filter) {
-                        Completed -> getCompletedUseCase.getLocalCompleted(PROGRESS_SECTION_LIMIT)
+                        UpToDate -> getCompletedUseCase.getLocalCompleted(PROGRESS_SECTION_LIMIT, ended = false)
+                        Ended -> getCompletedUseCase.getLocalCompleted(PROGRESS_SECTION_LIMIT, ended = true)
                         InProgress -> getWatchingUseCase.getLocalWatching(PROGRESS_SECTION_LIMIT)
                         Dropped -> getDroppedUseCase.getLocalDropped(PROGRESS_SECTION_LIMIT)
                     }
@@ -118,7 +121,18 @@ internal class ProfileProgressViewModel(
 
                 itemsState.update {
                     when (filter) {
-                        Completed -> getCompletedUseCase.getCompleted(page = 1, limit = PROGRESS_SECTION_LIMIT)
+                        UpToDate ->
+                            getCompletedUseCase
+                                .getCompleted(page = 1, limit = PROGRESS_SECTION_LIMIT)
+                                .pageOfEnded(ended = false)
+                                .items
+
+                        Ended ->
+                            getCompletedUseCase
+                                .getCompleted(page = 1, limit = PROGRESS_SECTION_LIMIT)
+                                .pageOfEnded(ended = true)
+                                .items
+
                         InProgress -> getWatchingUseCase.getWatching(page = 1, limit = PROGRESS_SECTION_LIMIT)
                         Dropped -> getDroppedUseCase.getDropped(page = 1, limit = PROGRESS_SECTION_LIMIT)
                     }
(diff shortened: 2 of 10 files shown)


michaldrabik:
Hi and thanks for the PR,

michaldrabik:
For this one specifically this is something we would need to introduce on the backend/api side first but I agree with the idea in general.

matheus-souza:
Thanks for the steer - I took it to the web first.

**trakt/trakt-web#3210** (feature request) and **trakt/trakt-web#3211** (implementation) now cover the split there.

Checking the endpoint against my account while writing that up settled the API question: **the split needs no API change.** `up_next_nitro` returns the full show object today, `status` included, even though the contract documents `status` as only available with `extended=full`:

```
GET /sync/progress/up_next_nitro?intent=completed&limit=100
  16 shows -> 10 ended, 6 returning series, 0 without a status
```

The web already maps that value onto `ProgressEntry.show.status` via `mapToShowEntry` and simply does not use it, so the web PR is a pure client-side change sharing one query between both filters.

**On this PR:** I have reduced it to the split alone - the counts and the settings toggle are gone, since those were the part that would have needed server-side work. What is left is just the filter.

matheus-souza:
Reduced this to the split alone, as promised above.

**26 files / +445 -204** → **10 files / +112 -24**. Gone: the per-chip counts, the `Show Progress Counts` setting, `ProgressCountsUseCase`, `ProgressOverview`, `LoadAllPages` and the full-bucket paging it drove, plus the two new strings. The original per-page pagination from `main` is back.

What is left needs no API change: both filters read the same `intent=completed` bucket and split it on the `status` the endpoint already returns.

One thing worth flagging, since it was easy to get wrong. Partitioning a page client-side shrinks it, and the pager decided `hasMoreData` from the size of what it received. Left alone, a page of 100 completed shows filtering down to 8 ended ones would have stopped pagination early. `ProgressPage` now carries the count the endpoint actually returned alongside the filtered items, and the pager reads that instead.

Same change is up for the web in trakt/trakt-web#3211, following the parity guidance.

matheus-souza:
Opened the API-side report for the counts blocker: trakt/trakt-api#926.

Scoped it a bit further while writing it up: both `up_next` and `up_next_nitro` emit the same hardcoded `1337`, and `/users/me/history` on the same account returns a coherent `item-count`/`page-count` pair - so the pagination layer itself is fine and this looks specific to up next.

Nothing to change here; this PR stays as the split alone.

matheus-souza:
Update on the parity condition: the web counterpart has landed.

**trakt/trakt-web#3211** was approved by @Marius-TV and merged into `main` on Sep 8. It ships the same split this PR does, with the same shape:

- one shared `intent=completed` query, partitioned client-side on the show status
- **Up to date** for still-airing shows you are caught up on, **Ended** for `ended` / `canceled`
- a `hasEnded()` util next to the other media utils, so the classification is one place
- **no API change** - which also answers the backend concern raised above: the status already comes back on the show object, so nothing new was needed server-side

So the web is now the reference implementation and this PR mirrors it rather than leading it. Naming matches (`Up to date` / `Ended`), the partition rule matches (`ended` or `canceled`, unknown status counts as still airing), and the scope matches - split only, no counts, no settings toggle.

Happy to rebase or adjust anything here if you would rather this track the merged web version more closely in any particular detail.

matheus-souza:
@michaldrabik the parity condition you raised is now cleared, and this PR is ready on the Android side.

**Web:** trakt/trakt-web#3211 was approved by @Marius-TV and merged into `main` on Sep 8. It also answers the backend point - the split needed **no API change**, because the show status already comes back on the up next response, so both filters read the same completed bucket and partition it client-side.

**Android:** this PR mirrors that implementation - same naming (`Up to date` / `Ended`), same partition rule (`ended` or `canceled`, unknown status counts as still airing), same scope (split only: no counts, no settings toggle, no API change).

Just rebased onto `main` (it was 34 commits behind). No conflicts, and no upstream commit has touched any of the 10 files this changes. Verified after the rebase:

```
./gradlew :app:compileInternalDebugKotlin    BUILD SUCCESSFUL
ktlint 1.7.1                                 clean on every changed file
```

Happy to adjust anything if you would rather it track the merged web version more closely in some detail.

PR #356, opened 2026-08-27
feat(app): open watch again from the watched check

### LINKED ISSUE #355: feat: open watch again from the watched check, and give seasons a context menu

Marking something as watched a second time is harder than it should be, and the most obvious control does the opposite of what it looks like.

On a movie, show or episode that is already watched, the purple check in the action bar is the natural "I watched this again" affordance. Tapping it opens the **Remove from History** confirmation instead. A filled check reads as "done", so tapping it again reads as "done again" — not "undo everything". Re-watching is possible today, but only by opening the overflow menu and picking **Watch again**, which is a second tap behind a menu.

Seasons are worse:

- The overflow on a season poster skips any menu and goes straight to the destructive **Remove from History** confirmation, unlike shows, movies and episodes, which all get a proper context menu first.
- The white watched chip on a season poster, and on watched episode rows inside a season, is decorative. Tapping it does nothing at all.
- **Watch again** is missing entirely from the overflow of a watched episode row inside a season — it only offers **Remove from History**.

There is also a bug underneath: `addToWatched(season)` filters out episodes that are already watched. On a fully watched season that resolves to an empty batch, so re-watching a season silently sends nothing.

Image

Image

Make the watched check mean "watched again" everywhere, and keep removal where destructive actions belong — behind the overflow menu.

1. **Movie / show / episode details** — tapping the check while watched opens the same watch again flow the overflow menu already uses (date selection; for shows, the existing hold-to-confirm first). Removal stays in the overflow.
2. **Season overflow** — opens a context sheet like every other media type, with **Watch again** and **Remove from History**, instead of jumping straight to the destructive confirmation.
3. **Season watched chip** — tapping it starts the same watch again flow.
4. **Watched episode row inside a season** — the overflow gains **Watch again**; tapping the watched chip on the episode cover opens that same menu.
5. **Season re-watch fix** — a completed season re-watches every released episode; a partially watched season still fills only the gaps.

The user setting that disables multiple plays keeps its current meaning: with it on, the check still opens the removal confirmation and **Watch again** stays hidden, exactly as the overflow menu already behaves.

No new strings are needed — this reuses `button_text_watch_again`, `button_text_track` and `button_text_remove_from_history`, all already translated.

- **Leaving the check as remove and relying on the overflow menu.** This is the current behaviour. It keeps re-watching two taps deep and leaves the check reading as the opposite of what it does.
- **A long press on the check for watch again.** Undiscoverable, and the long press on media cards is already taken by the context menu.
- **A real season-level `/sync/history` add.** The API accepts a `seasons[]` payload, so a season could be added in one request instead of fanning out to episode ids. That is a larger data-layer change and can be done separately; the fan-out path already exists and works.
- **Making the watched chip clickable app-wide.** Rejected — the chip appears on every media card in home, discover and lists, where a tap should keep navigating. The change is opt-in per call site.

### Additional context

Image

Image

matheus-souza:
Closes #355

Mirrored on the web in trakt/trakt-web#3213 (issue) and trakt/trakt-web#3214 (implementation), where `TrackAction` had the identical behaviour. Only the first part of #355 ports there - the web has no season action bar, so the season context sheet, the clickable watched chips, the episode rows inside a season and the `addToWatched(season)` fix stay specific to this PR.

Tapping the check on an already watched title only offered to remove it from history, which reads as "undo" rather than "I watched it again". Re-watching worked, but only from the overflow menu. This makes the check mean "watched again" everywhere and keeps removal behind the overflow, where destructive actions belong.

## Changes

**Details screens (movie / show / episode)** — `onTrackClick` on a watched title now routes to the same flow the overflow's **Watch again** already used: `DateSelectionSheet` for movies and episodes, the existing hold-to-confirm then date for shows. Three lambdas changed; no ViewModel or state changes, since `state.user` was already there.

**New `SeasonContextSheet`** — the season overflow and long press used to branch straight into `RemoveConfirmationSheet` / `ConfirmationSheet`. They now open a sheet like every other media type, with **Watch again** (or **Track**) and **Remove from History**. It mirrors `EpisodeContextSheet` / `EpisodeContextView`, needs no ViewModel, and is wired into both season surfaces — the full-screen `AllShowSeasonsScreen` and the section embedded in show details. There is no **View history** item because no per-season history screen exists.

**Season re-watch fix** — `addToWatched(season)` filtered on `!it.isWatched`, so a completed season resolved to an empty batch and re-watching silently sent nothing. That decision moved into a derived `ShowSeasons.selectedSeasonEpisodesToTrack`: every released episode when the season is complete, only the gaps otherwise. This also removed the filter duplicated across two ViewModels and two confirmation sheets, and it is what makes the confirmation read "24 episodes" instead of "0".

**Watched chips are no longer inert** — `VerticalMediaCard` and `PanelHorizontalMediaCard` gained `onCollectionChipClick: (() -> Unit)? = null`. It is opt-in and defaults to `null`, so every other call site (home, discover, lists) is untouched and its chips keep passing taps through to the card. Only the season rail and the episode rows inside a season pass a callback: the season chip starts the watch again flow, the episode chip opens that row's context menu. A clickable chip carries 8dp of hit slop, compensated in `translationY` so the pill does not move.

**Episode rows inside a season** — `EpisodeContextView` gained **Watch again** above **Remove from History** when watched, matching the episode details overflow.

`watchOnlyOnce` keeps its meaning throughout: with it on, the check still opens the removal confirmation and **Watch again** stays hidden, reusing the same gate the overflow menus already had.

No new strings — this reuses `button_text_watch_again`, `button_text_track` and `button_text_remove_from_history`. Nothing changed in `:common`, `:resources` or the networking layer; `POST /sync/history` already appends a play per add.

## Testing

`./gradlew :app:assemblePlaystoreDebug` and ktlint 1.7.1 both pass. Verified on an emulator against a real account:

| Scenario | Result |
| --- | --- |
| Watched movie, tap check | Opens date selection |
| Watched show, tap check | Opens "mark all N episodes", then date |
| Watched episode, tap check | Opens date selection |
| Season overflow | Opens the new sheet with Watch again + Remove from History |
| Season Watch again on a complete season | Confirms 24 episodes, not 0 |
| Season watched chip | Starts the watch again flow |
| Partially watched season | Still counts only the gaps |
| Episode row overflow, watched | Watch again + Remove from History |
| Episode row watched chip | Opens that row's context menu |
| Other media cards | Chips still inert, card tap still navigates |

End to end: after re-watching a show from the check, its history shows two full sets of plays and the episode header reads "WATCHED • 2".

05-depois-show-check-watch-again

07-depois-history-two-plays

11-depois-season-context-sheet

12-depois-season-24-episodes

15-depois-episode-chip-menu

## Note for reviewers

`isSelectedSeasonWatched` counts unaired episodes, so a returning show's current season never reads as fully watched and its sheet shows **Track** rather than **Watch again**. That is pre-existing behaviour — the same predicate drove the old long press branch — and I left it alone, but it does mean season re-watch only surfaces on completed seasons. Happy to change it to ignore unreleased episodes if you would prefer that.

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/core/summary/episodes/EpisodeDetailsScreen.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/summary/episodes/EpisodeDetailsScreen.kt
@@ -149,7 +149,8 @@ internal fun EpisodeDetailsScreen(
         onShowClick = onShowClick,
         onEpisodeClick = viewModel::navigateToEpisode,
         onTrackClick = {
-            if (state.episodeProgress?.watched == true) {
+            val watched = state.episodeProgress?.watched == true
+            if (watched && state.user?.settings?.watchOnlyOnce == true) {
                 confirmRemoveWatchedSheet = true
             } else {
                 dateSheet = true
--- a/app/src/main/java/tv/trakt/trakt/core/summary/movies/MovieDetailsScreen.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/summary/movies/MovieDetailsScreen.kt
@@ -168,7 +168,8 @@ internal fun MovieDetailsScreen(
             }
         },
         onTrackClick = {
-            if ((state.movieProgress?.plays ?: 0) > 0) {
+            val watched = (state.movieProgress?.plays ?: 0) > 0
+            if (watched && state.user?.settings?.watchOnlyOnce == true) {
                 confirmRemoveWatchedSheet = true
             } else {
                 dateSheet = true
--- a/app/src/main/java/tv/trakt/trakt/core/summary/shows/ShowDetailsScreen.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/summary/shows/ShowDetailsScreen.kt
@@ -183,7 +183,8 @@ internal fun ShowDetailsScreen(
             }
         },
         onTrackClick = {
-            if (state.showProgress?.isWatched == true) {
+            val watched = state.showProgress?.isWatched == true
+            if (watched && state.user?.settings?.watchOnlyOnce == true) {
                 confirmRemoveWatchedSheet = true
             } else {
                 confirmAddWatchedSheet = true
--- a/app/src/main/java/tv/trakt/trakt/core/summary/shows/features/context/seasons/SeasonContextSheet.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/summary/shows/features/context/seasons/SeasonContextSheet.kt
@@ -1,57 +1,54 @@
 @file:OptIn(ExperimentalMaterial3Api::class)
 
-package tv.trakt.trakt.core.summary.shows.features.context.episodes
+package tv.trakt.trakt.core.summary.shows.features.context.seasons
 
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.SheetState
-import androidx.compose.material3.rememberModalBottomSheetState
+import androidx.compose.material3.SheetValue.Expanded
+import androidx.compose.material3.SheetValue.Hidden
+import androidx.compose.material3.rememberBottomSheetState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
 import kotlinx.coroutines.launch
-import tv.trakt.trakt.core.summary.shows.features.seasons.model.EpisodeItem
+import tv.trakt.trakt.core.summary.shows.features.seasons.model.SeasonItem
 import tv.trakt.trakt.ui.components.TraktBottomSheet
 
 @Composable
-internal fun EpisodeContextSheet(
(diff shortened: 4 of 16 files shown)


matheus-souza:
Following the parity guidance, the portable part of this is now up for the web:

- trakt/trakt-web#3213 - feature request
- trakt/trakt-web#3214 - implementation

The web has the identical behaviour: `TrackAction` confirms removal when watched, and **Watch again** sits in the overflow. The fix there mirrors this PR - the check opens the same date drawer, removal moves into the overflow, and `watch_only_once` keeps its meaning.

Scope note: only part 1 of #355 ports. The web has no season action bar, and the only two `TrackAction` call sites are the media and episode bars, so the season context sheet, the clickable watched chips, the episode rows inside a season and the `addToWatched(season)` fix have no equivalent surface there. Those stay specific to this PR.

One thing the web trip turned up that is worth mentioning here: `MarkAsWatchedButtonIntlProvider.label` ignored `isRewatching`, so the overflow's **Watch again** announced itself to a screen reader as "Remove from your History". Fixed in the web PR. Worth a look at whether the Android content descriptions have the same gap.

PR #360, opened 2026-08-27
feat(app): offer watched until here when a check skips episodes

### LINKED ISSUE #359: feat: offer watched until here when a check skips episodes

Checking an episode in the seasons list only ever marks that one episode, even when the check skips over episodes that were never watched.

Say the last thing marked on a show is S02E06 and I tap the check on S02E10. E07, E08 and E09 stay unwatched, quietly. Nothing in the UI says so. Up next now points at E07 while the show has a play on E10, the season progress is full of holes, and the only way to notice is to scroll back through the row and count the missing chips.

**Watched until here** already exists and fixes exactly this: it fills every gap from the start of the show up to the episode picked. But it lives behind the episode overflow menu, and only shows up for an episode that is *not* watched yet. Someone using the check as their normal "I watched this" gesture never crosses paths with it.

Image

Keep the check doing what it does today, and when it leaves released episodes behind, follow up by offering the action that fills them.

1. Tapping the check marks the picked episode, exactly as now.
2. If there were released, unwatched episodes before it, a prompt appears: **Watched until here** / **Cancel**.
3. Confirming opens the existing **Watched until here** sheet (Just now / Release date / Other date), unchanged.
4. Cancelling leaves just the episode that was checked. Same outcome as today.

Detail worth stating:

- The gap covers previous seasons too, not only the current one, because that is what **Watched until here** actually writes. Specials are left out, matching `GetWatchedUntilEpisodesUseCase`.
- The prompt is decided from state the screen already has (per season watched counts plus the selected season episode list), so a check that skips nothing costs no extra request.
- A re-watch does not prompt.
- Same behaviour on the show details seasons row and on the All seasons screen. Both already host the **Watched until here** sheet.
- Since the picked episode is marked first, the sheet that follows counts only the gaps and leaves the anchor alone, so it never gets a second play.

Image

Image

No new strings. It reuses `button_text_watched_until_here`, `button_label_watched_until_here` and `button_text_cancel`, all already translated.

- **Leave it as is.** The gap stays invisible and the feature that fixes it stays undiscoverable to the people who need it most.
- **Have the check itself fill the gap silently.** Too aggressive. Skipping ahead on purpose is a real thing people do, and marking episodes they did not watch is not something to do without asking.
- **Show the entry as a permanent button on the episode row.** More clutter on every row for something that only matters at the moment a gap is created.
- **Re-open the episode overflow menu after the check.** It is a literal "context menu", but at that point the episode is watched, so the menu would be showing Remove from History next to the one action actually meant. A two-button prompt says the thing and gets out of the way.

### Additional context

I have this implemented and tested against a real account on an emulator, and I am opening a PR alongside this issue.

Image

Image

matheus-souza:
## Problem

Checking an episode in the seasons list marks only that episode. Skip ahead and every episode between the last watched one and the picked one stays unwatched, with nothing saying so: up next points at the gap, the season progress has holes, and the show quietly drifts out of sync.

**Watched until here** already fills exactly those gaps, but it sits behind the episode overflow menu and only for an unwatched episode, so the people using the check as their normal gesture never find it.

01-setup-lacuna-antes-do-check

## Solution

The check keeps its behaviour. When it leaves released episodes behind, a prompt follows offering **Watched until here**; confirming opens the existing sheet, cancelling leaves just the episode that was checked.

- The gap spans previous seasons as well as the current one, because that is what **Watched until here** writes. Specials are excluded, matching `GetWatchedUntilEpisodesUseCase`.
- It is computed from state the screen already holds (per season watched counts plus the selected season episode list), so a check that skips nothing costs no extra request.
- A re-watch does not prompt.
- Wired on both screens that already host the sheet: the show details seasons row and All seasons.
- The anchor is marked before the sheet opens, so the sheet counts only the gaps and never gives the anchor a second play.

02-prompt-watched-until-here

03-sheet-3-episodios-ancora-fora

## Files

- `app/.../seasons/model/ShowSeasons.kt` - new `skippedEpisodesBefore(episode)`, pure, no I/O
- `app/.../seasons/{ShowSeasonsState,ShowSeasonsViewModel}.kt` - `watchedUntilPrompt` field, set on a successful add, cleared by `clearWatchedUntilPrompt()`
- `app/.../seasons/ShowSeasonsView.kt` - `LaunchedEffect` bridging that field to a `ConfirmationSheet`
- `app/.../seasons/all/{AllShowSeasonsState,AllShowSeasonsViewModel,AllShowSeasonsScreen,AllShowSeasonsSheets}.kt` - the same on All seasons

The sheet state on both screens now holds an `Episode?` instead of an `EpisodeItem?`, since only the episode was ever read out of it.

Nothing changes in `:common`, `:resources`, networking, DI or the OpenAPI client.

## i18n

No new keys. Reuses `button_text_watched_until_here` (title and confirm), `button_label_watched_until_here` (message, already takes the episode title) and `button_text_cancel`, all translated across the 30 locales.

## Validation

Emulator API 31, real account, against `upstream/main`. `:app:assemblePlaystoreDebug` and ktlint 1.7.1 both clean.

| # | Scenario | Result | Screenshot |
| --- | --- | --- | --- |
| 1 | S02E06 last watched, check S02E10 | E10 marked, prompt shown | 02-prompt-watched-until-here |
| 2 | Confirm the prompt | Sheet lists "3 episodes" (E07, E08, E09); anchor excluded | 03-sheet-3-episodios-ancora-fora |
| 3 | Confirm the sheet | E07-E09 marked, gap closed | 04-lacuna-preenchida |
| 4 | Check the next episode, no gap | Marked, no prompt | 05-sem-lacuna-sem-prompt |
| 5 | Previous season incomplete, check S03E11 | Prompt shown from the earlier season's gap | 06-prompt-lacuna-temporada-anterior |
| 6 | Cancel the prompt | Only the checked episode stays marked | 07-cancelar-mantem-so-o-ancora |
| 7 | All seasons, overflow > Track > date | Same prompt after the date is applied | 08-all-seasons-track-menu-prompt |

Not exercised on device: specials (guarded by `episode.season <= 0`) and the re-watch path, since the seasons list only renders a check for an episode that is not watched yet; the `!wasWatched` guard is there for the overflow **Track** entry.

Closes: #359

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/core/summary/shows/features/seasons/ShowSeasonsState.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/summary/shows/features/seasons/ShowSeasonsState.kt
@@ -3,6 +3,7 @@ package tv.trakt.trakt.core.summary.shows.features.seasons
 import androidx.compose.runtime.Immutable
 import tv.trakt.trakt.common.helpers.LoadingState
 import tv.trakt.trakt.common.helpers.StringResource
+import tv.trakt.trakt.common.model.Episode
 import tv.trakt.trakt.common.model.Show
 import tv.trakt.trakt.core.summary.shows.features.seasons.model.ShowSeasons
 
@@ -16,4 +17,5 @@ internal data class ShowSeasonsState(
     val info: StringResource? = null,
     val error: Exception? = null,
     val collapsed: Boolean? = null,
+    val watchedUntilPrompt: Episode? = null,
 )
--- a/app/src/main/java/tv/trakt/trakt/core/summary/shows/features/seasons/ShowSeasonsView.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/summary/shows/features/seasons/ShowSeasonsView.kt
@@ -114,7 +114,8 @@ internal fun ShowSeasonsView(
     var confirmRemoveSeasonSheet by remember { mutableStateOf(false) }
 
     var episodeDateSheet by remember { mutableStateOf<EpisodeItem?>(null) }
-    var episodeWatchedUntilSheet by remember { mutableStateOf<EpisodeItem?>(null) }
+    var episodeWatchedUntilSheet by remember { mutableStateOf<Episode?>(null) }
+    var confirmWatchedUntilSheet by remember { mutableStateOf<Episode?>(null) }
     var episodeContextSheet by remember { mutableStateOf<EpisodeItem?>(null) }
     var seasonDateSheet by remember { mutableStateOf(false) }
 
@@ -161,7 +162,7 @@ internal fun ShowSeasonsView(
             episodeDateSheet = it
         },
         onWatchedUntilClick = {
-            episodeWatchedUntilSheet = it
+            episodeWatchedUntilSheet = it.episode
         },
         onRemoveClick = {
             confirmRemoveEpisodeSheet = it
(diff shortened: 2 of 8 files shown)


michaldrabik:
Hi and thanks for the PR,

matheus-souza:
Implemented on the web, as promised above:

- trakt/trakt-web#3217 - feature request
- trakt/trakt-web#3218 - implementation

The web had the same gap rather than the feature, so this was not a case of the app running ahead of the website.

One difference worth recording: on web the check opens the mark-as-watched drawer rather than marking directly, and the episode can also be marked from the overflow, so the prompt there keys off the episode becoming watched instead of off the click. The hook in this PR can stay as it is - just noting the two are not literally the same trigger.

PR #364, opened 2026-08-28
feat(app): reorder the progress filter chips by long press

### LINKED ISSUE #363: feat: let the progress filter chips be reordered by long press

The progress filters render in the order the enum declares them, so the row is always `Up to date`, `In progress`, `Dropped`, for everyone.

That order suits someone catching up on finished shows. It does not suit someone who opens the app to see what is still running: `In progress` is the only chip they ever want, and it always sits behind one they rarely consult. Every visit to Profile > Progress, and to the full progress screen behind the section title, costs an extra tap to get past a filter that is not theirs.

The selected filter already persists (`key_profile_progress_filter`), so the app is willing to remember a preference here. The order is simply not one of them.

Image

Let the chips be dragged into whatever order suits the person using them.

1. **A long press claims a chip** and slides straight into a drag, with the usual long press haptic. While the row is in this mode the chips rock gently, so it reads as editable rather than broken, and a plain horizontal swipe is enough to move another chip.
2. **The mode lets go on its own**: back, a tap on a chip, or a few seconds of no gesture. There is no bar to dismiss and nothing new on screen once it is over.
3. **A subtle one-off hint**, `Drag to reorder`, the first time the mode is entered, through the same `TutorialsManager` the rating bar swipe hint already uses.
4. **The order persists** next to the selected filter, in the same profile preferences store, and applies to both places the row appears: the Progress section and the full progress screen.

Tapping a chip keeps doing exactly what it does today. The order is presentation only, so no request is repeated when it changes.

Image

No new strings: the hint reuses `drawer_title_reorder_list`, already translated in every locale for the list reorder screen. No new dependency: the drag engine behind that screen is hand written and already almost axis agnostic.

- **A "move first" item in a context menu on each chip.** Discoverable, but it needs a menu on a 28dp pill and only expresses one of the moves.
- **A setting in Settings.** Further from the thing being ordered, and it is a preference nobody would go looking for.
- **Making `In progress` the default for everyone.** It trades one fixed order for another; whoever preferred `Up to date` first then pays the tap.
- **A dedicated reorder screen, like lists have.** Right for a long list with per item actions, far too heavy for three chips.

### Additional context

The `:tv` module has no progress section, so there is nothing to mirror there.

Image

matheus-souza:
Closes #363

The progress filters render in enum declaration order, so `In progress` always sits behind `Up to date` even for someone who only ever consults it. The selected filter already persists; the order was not configurable. This lets the chips be dragged into whatever order suits the person using them, in both places the row appears.

03-depois-in-progress-primeiro

## The gesture

A long press claims a chip and slides straight into the drag, so it is one continuous motion rather than press, release, press again. While the row is in reorder mode the chips rock gently and a plain horizontal swipe moves another one. The mode lets go on back, on a tap, or after a few seconds without a gesture.

Two things about the chip made this less obvious than it looks:

- The chip is already clickable, and a clickable consumes the down event, so the handle reads the gesture from the start (`awaitFirstDown(requireUnconsumed = false)`) rather than waiting for an unconsumed one.
- The clickable also sits closer to the layout node, so it sees the release before an outer modifier does. Once the handle has claimed the chip it therefore reads and consumes on `PointerEventPass.Initial`; otherwise a press and hold that never moved would still select the filter on the way up.

The handle is horizontal only, so it can sit inside the vertically scrolling progress screen without stealing its scroll, and it mirrors its delta under a right to left layout.

## Changes

**`:app` - the drag engine moves and gains an orientation.** `core/lists/features/reorder/ui/DragDropState.kt` moves to `ui/components/reorder/`, since it now has a second consumer and this avoids a profile to lists dependency. It was already axis agnostic apart from two things: `DraggableItem` gains `orientation` and `reversed` for the translation, and a new `horizontalDragHandle` carries the press and hold gesture above. `dragHandle` is unchanged, both new parameters default to the vertical behaviour, and `ListReorderScreen` changes by its imports alone.

**`:app` - the reorderable row.** `ui/components/chips/ReorderableFilterChipGroup.kt` is a `LazyRow` mirroring `FilterChipGroup`'s appearance, so it can animate a swap through `animateItem` and drive the engine. `FilterChipGroup` and `FilterChip` are untouched, so every other chip row in the app is byte for byte what it was. The group only holds the order for the length of a gesture and hands the result back on the drop.

**`:app` - the filters.** `ProgressFilters` takes an `order` and an `onReorder`, and owns the reorder mode itself: it is transient UI state, so it lives in the composable rather than the ViewModel. `GetProgressFilterOrderUseCase` sits beside `GetProgressFilterUseCase` in the same profile preferences store, under `key_profile_progress_filter_order`. It sanitises what it reads, dropping names that no longer resolve and appending any filter the stored value does not mention, so a preference written by an older build survives the enum changing instead of hiding a chip.

**`:app` - state.** `ProfileProgressState` and `AllProgressState` gain `filterOrder`; both ViewModels gain a `filterOrderState` source and a `setFilterOrder` that updates optimistically and persists. No reload: the order is presentation, not a query.

**`:common` - the hint.** `TutorialKey` gains `PROGRESS_FILTERS_REORDER`, so the one-off nudge rides on the manager the rating bar swipe hint already uses.

No new dependency and no new string. The hint reuses `drawer_title_reorder_list` ("Drag to reorder"), already translated in every locale for the list reorder screen.

## Testing

`./gradlew :app:assembleInternalDebug` and ktlint 1.7.1 both pass.

Verified on an emulator (API 31) against a real account:

| Step | Result |
| --- | --- |
| Long press a chip | Haptic, chips rock, hint appears the first time only |
| Long press and drag in one motion | Chip moves, row settles into the new order |
| Swipe a chip while the mode is on | Moves without another long press |
| Long press and release without moving | Enters the mode and does **not** select the filter |
| Tap a chip | Selects it and leaves the mode, list reloads as before |
| Back while in the mode | Leaves the mode, stays on the profile |
| Section title to the full progress screen | Same order, same selected filter |
| Reorder on the full progress screen | Works, and does not steal the vertical scroll |
| Kill and reopen the app | Order still in place, on both screens |
| Enter the mode again | Hint does not come back |

02-modo-reordenacao-dica

04-filtro-funciona-apos-reordenar

05-all-progress-mesma-ordem

06-all-progress-reordenado

07-ordem-persistida-apos-restart

Three caveats worth stating:

- **The list reorder screen was not exercised at runtime**, because the account I tested with has no personal lists. Its diff is its imports; `dragHandle` is unchanged and both new `DraggableItem` parameters default to what it was doing before, so it is inert by construction, but I did not put a finger on it.
- **RTL was not exercised at runtime.** The handle mirrors its delta and `DraggableItem` mirrors the translation, both driven by `LocalLayoutDirection`, but I could not force an RTL locale on this emulator to confirm it visually.
- **Reordering by accessibility action is not implemented.** The chips are still tappable and labelled as they were, but moving one needs the drag; custom actions would need "Move left" and "Move right" strings, which felt like a separate change. Happy to add it here if you would rather it shipped together.


michaldrabik:
Hi and thanks for the PR,

matheus-souza:
Thanks - checked this one against the rule too, and it lands differently from the others.

There is no equivalent on the web. The progress filters there are a `Toggler` with a fixed option list and no reordering of any kind, so this is new behaviour rather than the app catching up.

By your guidance that means it starts on the web side, so I will open the issue there and bring the implementation with it, as I did for trakt/trakt-web#3211, #3214 and #3216. I will link them here once they are up.

matheus-souza:
Implemented on the web, as promised above:

- trakt/trakt-web#3220 - feature request
- trakt/trakt-web#3221 - implementation

The web had no reordering at all, so this went in as new behaviour there rather than the app catching up. The gesture matches this PR - holding a chip slides into the drag - and the order persists next to the selected filter, which `useToggler` already stored.

PR #366, opened 2026-08-31
feature: hide recommendations

michaldrabik:
- add option to hide recommended show/movie


app/src/main/java/tv/trakt/trakt/core/sync/data/remote/shows/ShowsSyncApiClient.kt
```diff
@@ -169,6 +171,11 @@ internal class ShowsSyncApiClient(
         cacheMarker.invalidate()
     }
 
+    override suspend fun hideRecommendation(showId: TraktId) {
+        recommendationsApi.deleteRecommendationsShowsHide(id = showId.value)
```
michaldrabik: high

The generated `deleteRecommendationsShowsHide` does not accept a `Long` — regenerating the client from the committed spec (`./gradlew openApiGenerate`) produces `deleteRecommendationsShowsHide(id: DeleteRecommendationsMoviesHideIdParameter)`, where `DeleteRecommendationsMoviesHideIdParameter` is an empty class the generator emits for this path parameter's oneOf schema. A clean checkout will fail to compile this call, and even if it type-checked, the path would be built via the parameter's `toString()`. Since the generated endpoint is unusable as-is and the OpenAPI client must not be hand-edited, we should route the hide through `usersApi.postUsersHiddenAdd(section = "recommendations", …)` — the same pattern `dropShow` already uses in this file, and `POST /users/hidden/recommendations` is the documented equivalent of this endpoint.

```kotlin
override suspend fun hideRecommendation(showId: TraktId) {
    val request = PostUsersListsListAddRequest(
        shows = listOf(
            PostUsersListsListAddRequestShowsInner(
                ids = PostCheckinStartRequestOneOfOneOfEpisodeIds(
                    trakt = showId.value,
                    slug = null,
                    imdb = null,
                    tmdb = null,
                    tvdb = 0,
                ),
                title = "",
                year = 0,
            ),
        ),
    )
    usersApi.postUsersHiddenAdd(
        section = "recommendations",
        postUsersListsListAddRequest = request,
    )
    cacheMarker.invalidate()
}
```

app/src/main/java/tv/trakt/trakt/core/sync/data/remote/movies/MoviesSyncApiClient.kt
```diff
@@ -132,6 +136,11 @@ internal class MoviesSyncApiClient(
         cacheMarker.invalidate()
     }
 
+    override suspend fun hideRecommendation(movieId: TraktId) {
+        recommendationsApi.deleteRecommendationsMoviesHide(id = movieId.value)
```
michaldrabik: high

Same issue as the shows client: the regenerated `deleteRecommendationsMoviesHide` signature is `deleteRecommendationsMoviesHide(id: DeleteRecommendationsMoviesHideIdParameter)` — an empty generated class, not a `Long` — so this call breaks a clean `openApiGenerate` build. We should use `usersApi.postUsersHiddenAdd(section = "recommendations", …)` with a movies payload instead; that also gives the newly injected `usersApi` a purpose, and the `recommendationsApi` constructor parameter can then be removed.

```kotlin
override suspend fun hideRecommendation(movieId: TraktId) {
    val request = PostUsersListsListAddRequest(
        movies = listOf(
            PostUsersListsListAddRequestMoviesInner(
                ids = PostCheckinStartRequestOneOf1MovieIds(
                    trakt = movieId.value,
                    slug = null,
                    imdb = null,
                    tmdb = 0,
                ),
                title = "",
                year = 0,
            ),
        ),
    )
    usersApi.postUsersHiddenAdd(
        section = "recommendations",
        postUsersListsListAddRequest = request,
    )
    cacheMarker.invalidate()
}
```

app/src/main/java/tv/trakt/trakt/core/home/sections/recommended/HomeRecommendedViewModel.kt
```diff
@@ -118,6 +126,40 @@ internal class HomeRecommendedViewModel(
         }
     }
 
+    fun hideRecommendation(show: Show) {
```
michaldrabik, line 131: medium

The item is removed from `itemsState` before the suspend call runs, and the `catch` block only records the error. If the hide request fails, the card has already disappeared from the row, the show is still present in the API and the local recommended cache, and the user gets no feedback — the item silently reappears on the next refresh. Since this ViewModel already parks failures in `errorState` for its loaders, we should perform the remote hide first and only update the list once it succeeds (the same applies to the `hideRecommendation(movie: Movie)` overload below).

```kotlin
fun hideRecommendation(show: Show) {
    viewModelScope.launch {
        try {
            hideRecommendedShowUseCase.hideShow(show.ids.trakt)
            itemsState.update { items ->
                items
                    ?.filterNot { it is ShowItem && it.id == show.ids.trakt }
                    ?.toImmutableList()
            }
        } catch (error: Exception) {
            error.rethrowCancellation {
                errorState.update { error }
                Timber.recordError(error)
            }
        }
    }
}
```

app/src/main/java/tv/trakt/trakt/core/discover/sections/all/AllDiscoverViewModel.kt
```diff
@@ -266,6 +274,40 @@ internal class AllDiscoverViewModel(
         }
     }
 
+    fun hideRecommendation(show: Show) {
```
michaldrabik, line 279: medium

Same optimistic-removal concern as in `HomeRecommendedViewModel`: the list is updated before `hideShow` / `hideMovie` completes, and a failure is only logged — the row entry vanishes even though the recommendation was never hidden, then comes back on the next load with no explanation. We should hide first and mutate `itemsState` on success, letting the failure land in `errorState` as the rest of this ViewModel does.

```kotlin
fun hideRecommendation(show: Show) {
    viewModelScope.launch {
        try {
            hideRecommendedShowUseCase.hideShow(show.ids.trakt)
            itemsState.update { items ->
                items
                    ?.filterNot { it is ShowItem && it.id == show.ids.trakt }
                    ?.toImmutableList()
            }
        } catch (error: Exception) {
            error.rethrowCancellation {
                errorState.update { error }
                Timber.recordError(error)
            }
        }
    }
}
```

michaldrabik (commented):
## Code Review

This pull request introduces the ability to hide show and movie recommendations. It adds `HideRecommendedShowUseCase` and `HideRecommendedMovieUseCase` that call the recommendations hide endpoint and evict the item from the local recommended cache, wires new `hideRecommendation` handlers into `HomeRecommendedViewModel` and `AllDiscoverViewModel`, surfaces a confirmation sheet and a new "Hide recommendation" action in both context sheets, renames the `DiscoverSection` enum entries to PascalCase, and refreshes the watchlist button iconography.

Feedback on the changes focuses on the remote hide calls in the sync API clients, which pass a raw `Long` to generated endpoints whose parameter type is a broken empty class — a clean `openApiGenerate` build will not compile — recommending the same `usersApi.postUsersHiddenAdd(section = "recommendations", …)` pattern already used by `dropShow`; and on the ViewModel hide handlers, which remove the item from the list before the network call succeeds and swallow failures without restoring the item or surfacing an error.

PR #367, opened 2026-09-01
feature: why this recommendation

michaldrabik:
- adds Why This feature for recommended items

### DIFF
--- a/app/src/main/java/tv/trakt/trakt/core/home/sections/recommended/HomeRecommendedView.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/home/sections/recommended/HomeRecommendedView.kt
@@ -140,22 +140,18 @@ internal fun HomeRecommendedView(
     ShowContextSheet(
         show = contextShowSheet?.show,
         showRecommended = true,
+        showWhyThis = contextShowSheet?.sources?.isNotEmpty() == true,
         onHideRecommendation = { viewModel.hideRecommendation(it) },
-        onWhyThis = {
-            whyThisItem = contextShowSheet
-                ?.takeIf { it.sources.isNotEmpty() }
-        },
+        onWhyThis = { whyThisItem = contextShowSheet },
         onDismiss = { contextShowSheet = null },
     )
 
     MovieContextSheet(
         movie = contextMovieSheet?.movie,
         showRecommended = true,
+        showWhyThis = contextMovieSheet?.sources?.isNotEmpty() == true,
         onHideRecommendation = { viewModel.hideRecommendation(it) },
-        onWhyThis = {
-            whyThisItem = contextMovieSheet
-                ?.takeIf { it.sources.isNotEmpty() }
-        },
+        onWhyThis = { whyThisItem = contextMovieSheet },
         onDismiss = { contextMovieSheet = null },
     )
 
--- a/app/src/main/java/tv/trakt/trakt/core/home/sections/recommended/HomeRecommendedView.kt
+++ b/app/src/main/java/tv/trakt/trakt/core/home/sections/recommended/HomeRecommendedView.kt
@@ -63,10 +63,10 @@ import tv.trakt.trakt.common.model.MediaMode.Media
 import tv.trakt.trakt.common.model.Movie
 import tv.trakt.trakt.common.model.Show
 import tv.trakt.trakt.common.model.TraktId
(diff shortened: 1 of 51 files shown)


michaldrabik (commented):
## Code Review

This pull request introduces a "Why this recommendation" feature across the app. It adds new `/movies/recommendations` and `/shows/recommendations` V3 endpoints with typed request/response models, a dedicated `RecommendedItem` / `RecommendedSource` domain model for the home recommended section (replacing the reuse of `DiscoverItem`), a new `WhyThisSheet` bottom sheet that groups recommendation sources into "You watched", "You loved" and "You like" sections with a poster grid, and wires the sheet into both the home recommended row and the All Discover screen via the show/movie context sheets. It also relocates the recommended local data sources to a `data/` package and moves `LibraryFilter` under the profile feature.

Feedback on the changes focuses on the "Why this?" context action silently doing nothing when an item carries no sources — which is the guaranteed state for items restored from the pre-existing local cache — and on missing unit coverage for the new pure mapping helpers in `RecommendedSource`. As a side note, the PR title `feature: why this recommendation` does not match the Conventional Commits enum used for squash-merged titles; `feat(app): add why this recommendation sheet` would keep `main` history consistent.

app/src/main/java/tv/trakt/trakt/core/home/sections/recommended/HomeRecommendedView.kt
```diff
@@ -131,25 +131,38 @@ internal fun HomeRecommendedView(
...
     )
 
     ShowContextSheet(
-        show = contextShowSheet,
+        show = contextShowSheet?.show,
         showRecommended = true,
         onHideRecommendation = { viewModel.hideRecommendation(it) },
+        onWhyThis = {
+            whyThisItem = contextShowSheet
+                ?.takeIf { it.sources.isNotEmpty() }
```
michaldrabik: medium

When an item's `sources` list is empty, the "Why this?" button is still rendered in the context sheet (it is gated only on `showRecommended`), but tapping it silently does nothing because this callback assigns `null` to `whyThisItem`. This is the guaranteed state for every item restored from the existing on-device cache — entries persisted before this release decode with the new `sources` field defaulting to an empty list — so users will see a dead button until the next remote refresh. To avoid this, we should hide the action when there is nothing to show, for example by threading a visibility flag down to the context sheets. The same pattern applies to `AllDiscoverScreen`.

```kotlin
    ShowContextSheet(
        show = contextShowSheet?.show,
        showRecommended = true,
        showWhyThis = contextShowSheet?.sources?.isNotEmpty() == true,
        ...
    )
```

app/src/main/java/tv/trakt/trakt/core/home/sections/recommended/model/RecommendedSource.kt
```diff
@@ -0,0 +1,75 @@
...
+    companion object {
+        /**
+         * Merges distinct subgenres across sources, keeping their order.
+         */
+        fun List<RecommendedSource>.mergedSubgenres(): ImmutableList<Subgenre> =
+            flatMap { it.subgenres }
+                .distinctBy { it.slug }
+                .toImmutableList()
+
+        fun fromDto(dto: V3RecommendationSource): RecommendedSource =
```
michaldrabik, line 58: medium

This file introduces two pure helpers — `fromDto`, which parses the server's `favorite` / `activity` / `subgenre` strings into `Type` (with an `Unknown` fallback), and `mergedSubgenres`, which deduplicates subgenres by slug across sources — but the pull request ships no unit tests for either. The project standards list mappers and pure helpers as mandatory unit-coverage targets, and both functions have easily enumerable edge cases (unknown type value, duplicate slugs across sources, empty subgenre lists). We should add a `RecommendedSourceTest` covering these paths using plain fixtures.

```kotlin
@Test
fun `fromValue falls back to Unknown for unrecognised types`() {
    assertThat(RecommendedSource.Type.fromValue("editorial")).isEqualTo(Type.Unknown)
}

@Test
fun `mergedSubgenres dedupes by slug and keeps order`() {
    val merged = listOf(sourceWith("magic", "villain"), sourceWith("magic")).mergedSubgenres()
    assertThat(merged.map { it.slug }).containsExactly("magic", "villain").inOrder()
}

PR #370, opened 2026-09-03
feature: lists manage sheet multi-edit

michaldrabik:
- enable personal lists editing without closing sheet per each action


app/src/main/java/tv/trakt/trakt/core/summary/movies/MovieDetailsViewModel.kt
```diff
@@ -177,6 +172,16 @@ internal class MovieDetailsViewModel(
             .launchIn(viewModelScope)
     }
 
+    private fun observeLists() {
+        movieDetailsUpdates.observeUpdates(Source.Lists)
+            .distinctUntilChanged()
+            .debounce(200.milliseconds)
+            .onEach {
```
michaldrabik, line 179: high

`refreshLists()` now performs remote loads (`loadWatchlist()`, `loadLists()`) when the local caches are not yet populated. If one of these throws — for example, the device is offline when a list toggle notification arrives — the exception propagates out of `onEach`, which terminates the flow collection (so the screen stops reacting to further list updates) and surfaces as an uncaught exception in `viewModelScope`, crashing the app. The previous call sites always invoked `refreshLists()` inside a `try`/`catch`. We should wrap the call in the established `rethrowCancellation` error-handling pattern so a transient network failure degrades gracefully.

```kotlin
.onEach {
    try {
        refreshLists()
    } catch (error: Exception) {
        error.rethrowCancellation {
            Timber.recordError(error)
        }
    }
}
```

app/src/main/java/tv/trakt/trakt/core/summary/shows/ShowDetailsViewModel.kt
```diff
@@ -170,6 +166,16 @@ internal class ShowDetailsViewModel(
             .launchIn(viewModelScope)
     }
 
+    private fun observeLists() {
+        showDetailsUpdates.observeUpdates(Source.Lists)
+            .distinctUntilChanged()
+            .debounce(200.milliseconds)
+            .onEach {
```
michaldrabik, line 173: high

The same unguarded `refreshLists()` call exists in this `observeLists()` collector — the remote `loadWatchlist()` / `loadLists()` paths can throw, killing the collector and crashing the app when offline. We should apply the same `try` / `rethrowCancellation` wrapping as in `MovieDetailsViewModel`.

```kotlin
.onEach {
    try {
        refreshLists()
    } catch (error: Exception) {
        error.rethrowCancellation {
            Timber.recordError(error)
        }
    }
}
```

app/src/main/java/tv/trakt/trakt/core/summary/movies/features/context/lists/MovieDetailsListsViewModel.kt
```diff
@@ -79,24 +92,84 @@ internal class MovieDetailsListsViewModel(
         }
     }
 
-    fun isListed(listId: TraktId): Boolean {
+    fun toggleList(list: CustomListMinimal) {
```
michaldrabik, line 95: medium

This ViewModel (and its `ShowDetailsListsViewModel` mirror) is substantially rewritten with new behaviour that is easy to get subtly wrong — optimistic membership toggling, rollback on failure, the `toggling` guard against concurrent taps, and VIP-limit error routing — but the pull request adds no unit tests. ViewModel state emissions are part of the mandatory unit coverage for new and substantially rewritten code. We should add a `MovieDetailsListsViewModelTest` using hand-written fakes for the use cases, covering at least: a successful add/remove emitting the optimistic state, a failing toggle rolling membership back, and a second tap on an in-flight list being ignored.

```kotlin
class MovieDetailsListsViewModelTest {
    @Test
    fun `failed toggle rolls back optimistic membership`() = runTest {
        val viewModel = buildViewModel(addListItemUseCase = FailingAddListItemUseCase())
        viewModel.state.test {
            viewModel.toggleList(list)
            // optimistic add, then rollback on failure
            assertThat(awaitItem().movieLists).contains(list.id)
            assertThat(awaitItem().movieLists).doesNotContain(list.id)
        }
    }
}
```

michaldrabik (commented):
## Code Review

This pull request reworks the personal-lists manage sheet into a multi-edit experience: list membership toggling moves out of `MovieDetailsViewModel` / `ShowDetailsViewModel` into the sheet-scoped `MovieDetailsListsViewModel` / `ShowDetailsListsViewModel`, which now perform optimistic per-list toggles (with a `toggling` set, rollback on failure, and VIP-limit routing) without dismissing the sheet. A new `Lists` source on the details update buses notifies the parent ViewModels to re-run `refreshLists()`, the remove-confirmation step is dropped, and the previously unremembered random `koinViewModel` keys across 14 sheets are stabilised with `remember(...)`, with the pattern documented in `.agents/rules/state-management.md`.

Feedback on the changes focuses on error handling in the new `observeLists()` collectors — `refreshLists()` now performs remote loads whose failures are uncaught and would terminate the collector or crash the app when offline — and on the missing unit coverage for the substantially rewritten sheet ViewModels. As a side note, the PR title uses the type `feature:`, which is not in the allowed Conventional Commits enum; `feat(app): ...` would match the standard, since squash-merge carries the title onto `main`.

